PR #2, opened 2016-08-12
Pulp dev copr

seandst:
re https://pulp.plan.io/issues/1993#note-11


ansible/library/pulp_facts.py
```diff
@@ -48,14 +48,17 @@
         rpm_dependency_list += [rpm.split()[0] for rpm in stdout.splitlines()]
 
 # Remove any duplicates
-rpm_dependency_list = list(set(rpm_dependency_list))
+rpm_dependency_list = set(rpm_dependency_list)
+if 'python-twisted' in rpm_dependency_list:
+    rpm_dependency_list.remove('python-twisted')
+    rpm_dependency_list.add('python2-twisted')
```
seandst: We need to fix our specs to deal with this correctly, maybe this should be its own issue?

jeremycline: Just glancing at this I think this a bug in python-twisted's packaging. We should probably just have them fix it rather than work around it.

pulp-dev/pulp_dev/commands/pulp_dev/versions.py
```diff
@@ -0,0 +1,34 @@
+
+parser = argparse.ArgumentParser()
+help_text = "Specify which part of the version is being updated. "\
+            "Whichever part is specified will udpate all the other parts in accordance"\
```
jeremycline: s/udpate/update

pulp-dev/pulp_dev/git_workflow.py
```diff
@@ -0,0 +1,239 @@
...
+import subprocess
+import sys
+from itertools import tee
+
+from distutils.version import LooseVersion
+
+
+def promotion_chain(git_directory, commit_ref, parent_branch=None, skip_master=False):
+    """
+    For a given git repository direcotry & branch/tag, determine the promotion chain
```
jeremycline: s/direcotry/directory

pulp-dev/pulp_dev/releng_utils.py
```diff
@@ -0,0 +1,843 @@
...
+            next_stage = 'beta'
+        elif self.r_stage == 'beta':
+            next_stage = 'rc'
+        elif self.r_stage == 'rc':
+            next_stage = None  # GA release
+        else:
+            raise ValueError('Unknown self.r_stage: {}'.format(self.r_stage))
+        return next_stage
+
+    # XXX holy bejeebus this needs unit tests
```
jeremycline: heh

pulp-dev/pulp_dev/releng_utils.py
```diff
@@ -0,0 +1,843 @@
...
+        elif next_stage:
+            # always increment the minor release if advancing to a non-release stage
+            copy.r_minor += 1
+            if stage:
+                copy.r_stage = next_stage
+
+        return copy
+
+
+# py3: Remove this, use tempfile.TemporaryDirectory
```
jeremycline: Is this currently meant to be 2.7-3.5 compatible for now?

pulp-dev/pulp_dev/releng_utils.py
```diff
@@ -0,0 +1,843 @@
...
+
+def set_spec_field(field, specfile, new_value):
+    """
+    Set a specfile field.
+
+    This is not a great solution, but I can't find the tool to set a field
+    in a specfile.
+
+    :param specfile: path to the specfile
+    :param version:  a string to place in the version field
```
jeremycline: I think I'm responsible for this incorrect docblock

pulp-dev/pulp_dev/releng_utils.py
```diff
@@ -0,0 +1,843 @@
...
+            'mkdir -p "$(dirname "{dest}")" && '
+            'git archive --format=tar.gz --prefix={prefix}/ '
+            '--output={dest} HEAD &&'
+            'popd'
+        ).format(name=name, dest=destination, prefix=basename)
+
+        print('\nRunning "' + command + '"\n')
+        result = subprocess.check_output(command, shell=True)
+
+        # Record the commit used for the tarball in the configuration
```
jeremycline: Since the commit is being recorded above, this can go away, I think.

jeremycline:
The only thing I want to figure out what platforms aren't providing ``python-twisted`` as that's a packaging bug and I want to report it. Otherwise, looks like a good start.

pulp-dev/pulp_dev/releng_utils.py
seandst: The only reason I got py3 involved was part of an effort to get the "editable: True" to actually install pulp-dev as editable in the ansible tasks, which means I can just clean this up and go back to all-py3-all-the-time in this PR (which I'll do)

seandst:
I mentioned this inline, but I think I better understand the issues that were giving me troubles with py3 and I'll be going back and getting everything happy on py3.4+

seandst:
Made all the changes, but merged it to 3.0-dev since there are a few 3.0-specific changes.

PR #3, opened 2016-08-12
adding a script to set github branches as protected

mhrivnak:
re #2043
https://pulp.plan.io/issues/2043

### DIFF
--- a/scripts/protect-branches.py
+++ b/scripts/protect-branches.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python3
+
+
+import argparse
+import getpass
+import json
+import os
+import re
+import requests
+import sys
+import urllib
+
+
+# API token received from github. For the "scopes", I only selected the entire "repo" section.
+# https://github.com/settings/tokens
+TOKEN = getpass.getpass(prompt='API token from GitHub: ')
+
+
+# the Accept header tells github to let us use their experimental branch protection API
+headers = {
+    'Authorization': 'token {}'.format(TOKEN),
+    'Accept': 'application/vnd.github.loki-preview+json',
+}
+
+
+# the body to send when setting protection on a branch
+protection_body = json.dumps({'restrictions': None, 'required_status_checks': None})
+
+
+name_re = re.compile('^master|.*-dev|.*-release$')
+
+
+PULP_REPOS_URL = "https://api.github.com/repos/pulp/"
+
+
+repo_names = (
+    'crane',
+    'devel',
+    'nectar',
+    'pulp',
+    'pulp_docker',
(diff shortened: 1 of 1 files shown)


scripts/protect-branches.py
```diff
@@ -0,0 +1,97 @@
...
+import os
+import re
+import requests
+import sys
+import urllib
+
+
+# API token received from github. For the "scopes", I only selected the entire "repo" section.
+# https://github.com/settings/tokens
+TOKEN=""
```
bmbouter: Can we adjust this so that it reads from an environment variable. Since this file is under VCS and our edit would add a secret into the file it seems a little dangerous. A little comment saying set the GITHUB_TOKEN env variable would resolve this issue.

scripts/protect-branches.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
```
bmbouter: :+1:

bmbouter:
@mhrivnak Just one comment to switch the TOKEN storage to an env var. I do think that should be done. What do you think?

scripts/protect-branches.py
```diff
@@ -0,0 +1,97 @@
...
+    'pulp_python',
+    'pulp_rpm',
+)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Sets branches on GitHub to be protected. '
+                                     'Operates on master, *-release, and *-dev branches.')
+    parser.add_argument('-l', '--list', action='store_true',
+                        help='List branches that should be protected without making any changes.')
```
bmbouter: I tried this option and the explanation didn't match what it's behavior does. I'll suggest this replacement:

`List branches that should be protected but are not. This command makes no changes.`

scripts/protect-branches.py
mhrivnak: I thought about this too. It's a little inconvenient to set an environment variable without putting the value in your shell history, but it's doable. I could use ``getpass()`` perhaps which would be secure, but would make you enter it every time. Or both.

Part of my dilemma is that I wrote this little script to save time and thus tried to avoid feature creep, but I agree this is worth considering.

scripts/protect-branches.py
mhrivnak: I'll clarify it.

scripts/protect-branches.py
bmbouter: I think using getpass be good and I could help it the stdout the URL to get
the token.  that would be quick I think

bmbouter:
@mhrivnak Thanks for the change, it looks great.

PR #4, opened 2016-08-19
Merge pulp_packaging Ansible roles and playbooks

jeremycline:
I went with the pulp_packaging version of the ``lazy`` and ``pulp-certs`` roles, which are more general. I didn't attempt to de-duplicate anything beyond those overlapping roles, but this will at least have us all working in the same repository for Ansible.

### DIFF
--- a/ansible/dev-playbook.yml
+++ b/ansible/dev-playbook.yml
@@ -9,4 +9,4 @@
     - qpidd
     - dev
     - lazy
-    - pulp_ca
+    - pulp-certs
--- a/ansible/roles/lazy/files/squid-3.1.conf
+++ b/ansible/roles/lazy/files/squid-3.1.conf
@@ -2,7 +2,13 @@
 # matters in Squid's configuration; the configuration is applied top to bottom.
 
 # Listen on port 3128 in Accelerator (caching) mode.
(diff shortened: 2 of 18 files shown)


jeremycline:
@elyezer I'd appreciate it if you took a look at this to make sure it's reasonable.

elyezer:
This PR looks good to me. Something that I have in mind if we would have a separated repository for the playbooks or not. If we decided to have all playbook on pulp repo I will make sure that automation start using the playbooks in here instead of pulp_packaging and then we can drop the playbooks there.

Having the playbook here will help with the vagrant file because everything will be on a single repo. On the other hand having the playbook here will require automation to clone this repo instead of the pulp_packaging and possibly need to clone both in order to have files for the CI all together. Needing to clone two repos is not a problem but I am just saying what I identified so far to have in mind when migrating.

jeremycline:
I think I lean towards not including the playbooks here (except for maybe the one we use for development) since playbooks are tiny and deployment-specific.

jeremycline:
Okay, I've ripped out the non-Vagrant/dev playbooks

elyezer:
:+1:

PR #6, opened 2016-08-29
Ansible now looks for the pulp/packaging repo for specs

jeremycline:
In 3.0-dev branches, the spec files have been removed from the project
repositories. Ansible needs to look in the packaging repository to
gather the list of dependencies.

### DIFF
--- a/ansible/library/pulp_facts.py
+++ b/ansible/library/pulp_facts.py
@@ -4,8 +4,11 @@ This module gathers Pulp specific Ansible facts about the remote machine.
 """
 import json
 import os
+import platform
 import pwd
+import shutil
 import subprocess
+import tempfile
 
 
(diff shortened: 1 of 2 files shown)


ansible/library/pulp_facts.py
```diff
@@ -23,19 +24,40 @@
...
+platform_dist = platform.linux_distribution()
+if platform_dist[0].lower() == 'fedora':
+    # Maybe we'll support other platforms for development, but right now Fedora is
+    # all we work with.
+    branch = 'f' + str(platform_dist[1])
+else:
+    raise OSError(str(platform_dist) + ' is not a supported development platform')
+
+
+packaging_repo = os.path.join(user_homedir, 'devel/packaging/')
```
jeremycline: I'm on the fence about requiring the packaging repository to be checked out here. The alternative is to check it out in a temporary location and not rely on the user to clone it.

jeremycline:
Also note the dev guide will need an update if we decide to require the packaging repo.

PR #7, opened 2016-08-31
ansible: fix pulp_facts when selinux-policy not installed

rohanpm:
If selinux-policy package is not installed, rpmspec produces spurious
requirements for packages named "selinux-policy", "is", "not", "installed"
which breaks subsequent ansible tasks.  This is due to the macro in
pulp.spec:

  %define selinux_policyver %(rpm --qf "%%{version}-%%{release}" -q %%selinux-policy)

We know we'll need this package anyway, so just install it a bit
earlier to fix this.


jeremycline:
Thanks! :shipit:

PR #15, opened 2016-10-05
Starts pulp_streamer with other pulp services

werwty:
This is so plugins are loaded into the streamer.

closes #2272
https://pulp.plan.io/issues/2272


bmbouter:
@werwty Oh one thing, the commit should probably be redone to follow the commit length guidelines

werwty:
streamer is started with the pstart call in vagrant-setup.sh

PR #21, opened 2016-11-30
Overhaul ansible playbook and Vagrantfile for 3.0

seandst:
This isn't complete, largely because pulp's own 3.0-dev branch isn't
itself complete (we haven't yet removed all subpackages provided in
pulp 2, plugins aren't converted to pulp 3 yet, etc.).

The vagrant setup script has been destroyed, and its functionality
has been implemented in ansible. The ansible pulp_facts module has
been significantly improved, and should be much easier to maintain
by (and be in general less scary to) Pulp developers in general.

This commit still includes some TODO comments that I hope to get
figured out as part of this PR review, even if it's just to know what
follow-up tasks we need to make in Redmine.

re #2413
https://pulp.plan.io/issues/2413


seandst:
I'm still writing the docs for this, but if you want to test, this is a brief summary of what will be in the dev setup guide when it comes to the ansible playbook.

# ansible/python version

python 3.5+ and ansible 2.2+ are required for this to work. Both are currently available in fedora 24+. The playbook will warn you if this requirement is not met, but not fail. If it becomes a problem, we can alter the warning task to fail the playbook.

# With Vagrant

```
vagrant up # this should work, for starters
vagrant ssh
workon pulp
python manage.py runserver # ...or whatever you want, manage.py should work
```

# Without Vagrant

If you aren't using Vagrant, there's more work to do:

- create a vm using fedora 24 (or other fedorae at your own risk)
- ensure that the unprivileged user you set up in this vm has full sudo privs (`(ALL) NOPASSWD: ALL`)
- clone all the git repositories you want to work with into a single directory in the unprivileged user's homedir, checked out to the correct branches
- invoke the ansible playbook

ansible playbook invocation requires setting the `ansible_python_interpreter` to `/usr/bin/python3`. This playbook has not been and should not be tested using python 2. Also, if you checked your code out into a different directory than the vagrant default of `~/devel`, you'll need to set `pulp_devel_dir` to the absolute path on the VM to your code dir.

Manual playbook invocation, then, looks something like `ansible-playbook -e ansible_python_interpreter=/usr/bin/python3 -e pulp_devel_dir=/home/pulp/code -i inventory ansible/dev-playbook.yml`

Note that this only works with pulp on 3.0-dev. It does setup virtualenvs for the plugins, but since no plugin works with 3.0-dev yet, I didn't add tasks to set that up since they'll just fail.

seandst:
The final dev task, which is supposed to migrate the db, isn't. I'm currently tweaking that, and maybe breaking things, but I'll update here when it's working.

seandst:
I added a second (optional) commit to this to update the vagrant box to fedora 25.

ansible/library/pulp_facts.py
```diff
@@ -34,60 +24,169 @@
...
+    if platform_dist[0].lower() == 'fedora':
+        # right now Fedora is the only supported platform
+        branch = 'f' + str(platform_dist[1])
+    else:
+        msg = '{} is not a supported development platform'.format(platform_dist)
+        module.fail_json(msg=msg, name=platform_dist)
+
+    try:
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # TODO: Make this URL a module arg, pass it in to make it easy to use/test forks?
```
seandst: Can probably just leave the comment here, reword it a little so it isn't a question, and take out the TODO

ansible/roles/dev/tasks/main.yml
```diff
@@ -70,40 +56,138 @@
...
+
+  - name: Install Pulp Platform packages
+    command: "{{ pulp_venv_dir }}/pulp/bin/python setup.py develop"
+    args:
+      chdir: "{{ pulp_devel_dir }}/pulp/{{ item }}"
+    # Pulp 2 setup.py files still exist on 3.0-dev in platform, so we can't just find
+    # every setup.py and run develop on it with the virtualenv interpreter. This is also true
+    # for plugins, in which we'll likely want to install all plugin setup.py as well as
+    # all platform setup.pys for each plugin virtualenv.
+    # TODO: Make a redmine task to track this
```
seandst: I think https://pulp.plan.io/issues/278 covers this nicely, so I can just delete the TODO and add a reference to that issue

ansible/roles/dev/tasks/main.yml
```diff
@@ -11,35 +10,24 @@
...
-      regexp: '^vagrant'
-      line: 'vagrant ALL=(ALL) NOPASSWD: ALL'
-      validate: 'visudo -cf %s'
-      mode: 0440
-  when: vagrant_nopasswd.stat.exists
-  name: Rejigger vagrant sudoers file
-
+  # TODO: It's not clear how this list was created, it would be great
+  #       to know why we install all of these, and (if possible) get them
+  #       captured as requirements in the project itself.
```
seandst: In looking at this, a lot of these appear to be either packages needed to install python packages, python packages themselves, or developer-related things. I think I can probably sift through this, clean it up, and get a little more detail into the task name/comments so that this list isn't such a mystery.

bmbouter: That would be great!

bmbouter:
Does it make sense to pull these sections out of the playbook?

```
TASK [mongo : Install mongo packages] ******************************************
changed: [dev] => (item=[u'mongodb', u'mongodb-server'])

TASK [mongo : Disable MongoDB journal] *****************************************
ok: [dev]

TASK [mongo : Start and enable mongo services] *********************************
changed: [dev]
```

bmbouter:
Also what should we do about RabbitMQ support? It would be great for Qpid to be the default and RabbitMQ to be an option (which would not install Qpid or its client packages). Then we could run the playbook with a special option which would be useful for our dev environments which currently require RabbitMQ. We could make this a separate task also since this could be its own piece of work.

seandst:
> TASK [mongo : Install mongo packages] ****************************************** changed: [dev] => (item=[u'mongodb', u'mongodb-server']) TASK [mongo : Disable MongoDB journal] ***************************************** ok: [dev] TASK [mongo : Start and enable mongo services] ********************************* changed: [dev]

Yeah, I think we still need it. There's a comment in the tasks file for that role explaining why it still exists:
https://github.com/pulp/devel/blob/6d01fd93d64e68b61ff93fbbe4f8276ed0ee1ce0/ansible/roles/mongo/tasks/main.yml

bmbouter:
I think the systemd units are not being loaded.

```
[vagrant@dev ~]$ pstatus
Unit goferd.service could not be found.
Unit pulp_workers.service could not be found.
Unit pulp_celerybeat.service could not be found.
Unit pulp_resource_manager.service could not be found.
Unit pulp_streamer.service could not be found.
● httpd.service - The Apache HTTP Server
   Loaded: loaded (/usr/lib/systemd/system/httpd.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
```

When I try to run a worker interactively I switch to pulp with `workon pulp` and then run:

```
celery beat --app=pulp.tasking.celery_app:celery --scheduler=pulp.tasking.services.scheduler.Scheduler -l debug```

which runs pulp celerybeat as the `vagrant` user.  To run as similarly as users will it should be run as the `apache` user. When I run the typical interactive worker startup I run:

```
sudo -u apache celery beat --app=pulp.tasking.celery_app:celery --scheduler=pulp.tasking.services.scheduler.Scheduler -l debug
```

which says: `sudo: celery: command not found`.

**Comment by seandst** [comment]
I wasn't aware we actually had working unit files yet for pulp 3; probably I should be following that tasking work more closely.

**Comment by bmbouter** [comment]
They are in the wrong place, but they are up to date with the latest tasking system updates.

https://github.com/pulp/pulp/tree/3.0-dev/server/usr/lib/systemd/system

**Comment by bmbouter** [comment]
I think the systemd file should be moved to `app/system/<files_go_here/` and the vagrant file install them somehow from that location in the pulp repo.

**Comment by seandst** [comment]
That makes sense, I'll take care of it. Thanks!

**Comment by bmbouter** [comment]
Thanks @seandst

**Comment by seandst** [comment]
> That makes sense, I'll take care of it. Thanks!

This has turned out to be more work than I'm comfortable dealing with in the scope of this PR. As far as I can tell those unit files do not work in the current pulp 3 environment. They should live in their correct location and be working before we integrate them into the playbook, and that work should not block this PR. I can open up a redmine issue with specifics on what exactly is broken about them, if needed.

**Comment by bmbouter** [comment]
@seandst Can whatever you've learned about what needs to be done to fix these systemd files be written into a story?

**Comment by mhrivnak** [comment]
FWIW, prior to this PR, the systemd unit files for the worker services were working on 3.0-dev. Despite living in the wrong place, they were getting installed, and I was able to use them while getting the tasking system working.

```
(pulp) [vagrant@dev app]$ sudo systemctl start pulp_resource_manager
(pulp) [vagrant@dev app]$ sudo systemctl status pulp_resource_manager
● pulp_resource_manager.service - Pulp Resource Manager
   Loaded: loaded (/etc/systemd/system/pulp_resource_manager.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2016-12-07 15:11:18 UTC; 2s ago
 Main PID: 31980 (celery)
    Tasks: 2 (limit: 512)
   CGroup: /system.slice/pulp_resource_manager.service
           ├─31980 /bin/python3 /usr/bin/celery worker -A pulp.tasking.celery_app:celery -n resource_manager@%h -Q resource_manager -c 1 --events --umask 18 --pidfile=/var/run/pulp/resource_manager.pid --heartbe
           └─31986 /bin/python3 /usr/bin/celery worker -A pulp.tasking.celery_app:celery -n resource_manager@%h -Q resource_manager -c 1 --events --umask 18 --pidfile=/var/run/pulp/resource_manager.pid --heartbe

Dec 07 15:11:18 dev systemd[1]: Started Pulp Resource Manager.
Dec 07 15:11:20 dev celery[31980]: /usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/fixups/django.py:202: UserWarning: Using settings.DEBUG leads to a memory leak, never use this setting in product
Dec 07 15:11:20 dev celery[31980]:   warnings.warn('Using settings.DEBUG leads to a memory leak, never '
Dec 07 15:11:20 dev celery[31980]: [2016-12-07 15:11:20,175: WARNING/MainProcess] /usr/lib/python3.5/site-packages/celery-4.0.0-py3.5.egg/celery/fixups/django.py:202: UserWarning: Using settings.DEBUG leads to a
Dec 07 15:11:20 dev celery[31980]:   warnings.warn('Using settings.DEBUG leads to a memory leak, never '

(pulp) [vagrant@dev app]$ git status
On branch 3.0-dev
Your branch is up-to-date with 'upstream/3.0-dev'.
```

seandst:
I haven't been able to get the services working reliably in the dev environment installed via ansible, so I forked that off as a separate task:
https://pulp.plan.io/issues/2479

The service files are installed by the pulp 2 RPMs, so moving them will certainly break the dev install, but in my opinion needs to happen. We're lucky that they've continued to work, but that's a ticking time bomb that we should probably address directly. :)

PR #25, opened 2016-12-21
Problem: ansible is unable to install python-lxml

dkliban:
Solution: replace python-lxml with python2-lxml when gathering pulp facts


ansible/library/pulp_facts.py
```diff
@@ -50,6 +50,10 @@
 # Remove any duplicates
 rpm_dependency_list = list(set(rpm_dependency_list))
 
+# Replace python-lxml with python2-lxml
```
mhrivnak: It would be helpful to include a short explanation of why this is necessary, or perhaps just a link to the email thread.

PR #29, opened 2017-01-25
Vagrant: bump Fedora version when using docker

rohanpm:
Bring the major Fedora version used for docker in-line with the version
used with libvirt.  Both vagrant providers should be using the same base.

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -94,7 +94,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         override.vm.box = nil
         # Based on fedora official image, with vagrant's needs met:
         # https://github.com/rohanpm/docker-fedora-vagrant
-        d.image = 'rohanpm/fedora-vagrant:23'
+        d.image = 'rohanpm/fedora-vagrant:24'
 
         # use ssh for the sake of ansible
         d.has_ssh = true


rohanpm:
The purpose of this commit is only to achieve "both vagrant providers should be using the same base" mentioned in the commit message.  This is changing from having libvirt use F24 and docker use F23, to having both of them use F24.  Switching both of them to F25 seems like a different matter.  How does the Pulp project decide which Fedora version should be used as the base for the dev env?

seandst:
We use the latest fedora (when we remember to update the Vagrantfile). This one should be on fedora 25, so I'll do a quick test on that today and put up a separate PR to fix that.

Where is the motivation for "Both vagrant providers should be using the same base" coming from? It's a nice goal, but I think I'd rather we just use the latest Fedora as soon as possible.

seandst:
I think this looks good now, but it looks like jenkins is having some network problems. Once those clear up, I'd love to see the tests pass on all platforms before merging.

seandst:
...and I updated the wrong PR.

I'll be deleting these comments now. :D

seandst:
Based on that, I'm happy to merge this at f24, or f25. There's a link to the open PR for bumping the libvirt image to f25.

PR #30, opened 2017-01-27
Update Vagrant example to f25

seandst:
Tested this just now. We've already made this change on 3.0-dev, but we (really, I...) didn't bring that change to master, which is still on fedora 24. I just vagrant up'd with this and ran all tests, everything seems fine (which, again, was our experience on 3.0-dev).

Thanks to @rohanpm for pointing out this oversight in #29.


rohanpm:
@seandst do you want to simply update both the libvirt and docker cases in this commit?
It could even be extracted to a constant so they're always the same.

seandst:
The fedora and docker images have separate build systems, so I don't know that it's safe to do that. I don't know for sure that it *isn't* safe, either, but would prefer to err on the side of being explicit.

seandst:
Is the f25 docker image available? If so, I'll amend this commit to include it.

rohanpm:
Yes that is available already, it can be seen at https://hub.docker.com/r/rohanpm/fedora-vagrant/tags/ .

PR #34, opened 2017-02-09
Fixed a mistake which causes pulp_worker errors

dralley:


seandst:
Wow, that's subtle. I remember testing this and the VM came up seemingly with no issues :\

Thanks for the fix

PR #36, opened 2017-02-22
Vagrant stops installing pulp-server rpm

werwty:
This fix requires https://github.com/pulp/pulp/pull/2959 to be merged first

closes #2497
https://pulp.plan.io/issues/2497

### DIFF
--- a/ansible/roles/lazy/tasks/main.yml
+++ b/ansible/roles/lazy/tasks/main.yml
@@ -4,7 +4,6 @@
   with_items:
       - squid
       - httpd
-      - python-pulp-streamer
 
 - name: Install squid proxy configuration
   copy: src=squid.conf dest=/etc/squid/squid.conf
@@ -18,6 +17,3 @@
 
 - name: Start and enable Squid service
   service: name=squid state=started enabled=yes
-
-- name: Start and enable Pulp streamer service
-  service: name=pulp_streamer state=started enabled=yes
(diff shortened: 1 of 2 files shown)


werwty:
With this fix
`[vagrant@dev ~]$ sudo dnf list installed | grep pulp`
`[vagrant@dev ~]$`

bmbouter:
I tested this. I vagrant upped with this PR checked out and pulp/pulp master at 18f7c6ae1aa8c8a430e3fd2260ab11f8bdfd9078 which includes pulp/pulp#2959. Vagrant finished as expected and indeed I get the following output:

```
[vagrant@dev ~]$ sudo dnf list installed | grep pulp
python-kombu.noarch                    1:3.0.33-6.pulp.fc24            @pulp-nightlies
python-nectar.noarch                   1.5.3-1.fc24                    @pulp-nightlies
```

ansible/roles/lazy/tasks/main.yml
```diff
@@ -4,7 +4,6 @@
   with_items:
       - squid
       - httpd
-      - python-pulp-streamer
```
bmbouter: Don't we want this to run inside of vagrant though?

werwty: We don't need to install python-pulp-streamer via dnf when it's being installed with setuptools
https://github.com/pulp/pulp/blob/master/streamer/setup.py

This is also what was pulling in the unwanted pulp-server dependency before.

PR #37, opened 2017-03-13
Provisioning updates for Pulp 3

seandst:
- Example Vagrantfile updated with changes from pulp 2
- modified the gitignore to ignore different Vagrantfiles and
  all playbook retry files from ansible.
- Playbook updated to use the dev_requirements file intoduced in
  https://github.com/pulp/pulp/pull/2888

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,7 +93,7 @@ rel-eng/*
 !rel-eng/config
 
 # Vagrant
-Vagrantfile
+Vagrantfile*
 .vagrant
 .dnf-cache
-ansible/dev-playbook.retry
+ansible/*.retry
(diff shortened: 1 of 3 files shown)


dralley:
"vagrant up" fails because dev_requirements.txt is missing.  Did you forget to push it?

seandst:
In working with the new pulp_file plugin, I've discovered an annoyance with migrations that makes me think we just should make them until we decide the data model is good. I'm not sure if I'll open a separate issue for this, or just roll it up into this "fix 3.0-dev vagrant" omnibus to address this, though.

bmbouter:
@seandst regarding the migrations +1 to not committing them until later in the project. Feel free to fix/modify however you see fit.

seandst:
I'm finding a few things that I'd like to fix as part of the process of documenting how plugins should work with platform and be laid out. I think I'll merge this for now, and file/fix issues as appropriate. I think the migrations problem is something I should fix as part of my PRs to stub out the file plugin, since they're required for it to work, and the other issues we can fix in the normal sprint cycle. They're annoying, but not totally breaking.

PR #40, opened 2017-03-24
Added installation of 'attr' package to pulp-smash role

dralley:


Ichimonji10:
Pulp Smash uses the `getfattr` executable, which is provided by the "attr" package.

PR #43, opened 2017-04-06
Run pulp smash as documented by pulp-smash, with args

seandst:


seandst:
I haven't actually tested this on a pulp 2 vagrant env; I'm not sure if we were explicitly using py.test for a reason or not.

Ichimonji10:
`python -m pulp_smash` prints out usage information. It doesn't actually launch the test suite. `py.test path/to/code` is a good way to actually run the test suite.

I think I've seen some mis-understandings come from the Pulp Smash docs. I'll make a note to update what they say.

seandst:
Yeah, I'm planning to tear down my pulp 3 env today and actually get this tested and working for pulp 2.

There's no "WIP" label for the devel repo; this should be considered WIP.

seandst:
I have ideas and schemes about this, but at the moment not the means to make it happen. I'm going to open up a redmine task that explains what this is really about.

Ichimonji10:
Sounds good. For what its worth, I see some near-term changes to how the `python -m pulp_smash` tool works.

PR #44, opened 2017-04-12
Platform virtualenv installs pulp_file

seandst:
I'm been digging into figuring out why the platform virtualenv appeared
to be getting set up improperly, and had started down a scary path of
redoing how those virtualenvs get created. Then Austin asked me if the
'- plugin' item was left out of the platform virtualenv setup on
purpose. I'd just noticed that too, and discovered that this was the
actual cause of the virtualenv problems and I'd just missed it. Woot.

I was able to steal some of the virtualenv work I was doing, though, and
make the playbook also install pulp_file into the platform virtualenv if
the repo's cloned into the devel dir when the playbook runs, so that's
cool too.

### DIFF
--- a/ansible/library/pulp_facts.py
+++ b/ansible/library/pulp_facts.py
@@ -27,11 +27,12 @@ rpm_projects = [
 
 # plugin repository names, declared deparately from all repositories for easy filtering later
 plugins = [
-    'pulp_puppet',
-    'pulp_rpm',
     'pulp_docker',
+    'pulp_file',
     'pulp_ostree',
+    'pulp_puppet',
     'pulp_python',
+    'pulp_rpm',
 ]
 
 # all repository names, used first to make sure the platform repository is cloned in its configured
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -137,6 +137,21 @@
       - exceptions
       - tasking
 
+# We don't currently have a common method for installing plugins for development. In Pulp 2, it was
+# "run pulp-dev.py -I", so maybe that's what we'll do for pulp 3 as well. For now, since only the
+# file plugin works with "python setup.py develop", we'll just try that
+# exists, installing plugins to the platform virtualenv
+- name: Install Plugin packages into Platform Virtualenv
+  command: "{{ pulp_venv_dir }}/pulp/bin/python setup.py develop"
+  args:
+    chdir: "{{ pulp_devel_dir }}/{{ item }}"
+  # If we stick with running setup.py develop for all plugins, we'd want to do it for all available
+  # plugins. For now, it only works with pulp_file
+  # with_items: "{{ pulp_available_plugins }}"
+  with_items:
+      - pulp_file
+  when: "{{ [item] | issubset(pulp_available_plugins) }}"
+
 - block:
   - name: Install crane config
     template:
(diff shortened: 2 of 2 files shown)


ansible/library/pulp_facts.py
```diff
@@ -27,11 +27,12 @@
...
 # plugin repository names, declared deparately from all repositories for easy filtering later
 plugins = [
-    'pulp_puppet',
-    'pulp_rpm',
     'pulp_docker',
+    'pulp_file',
     'pulp_ostree',
+    'pulp_puppet',
     'pulp_python',
+    'pulp_rpm',
```
seandst: added pulp_file, other stuff is because sort

ansible/roles/dev/tasks/main.yml
```diff
@@ -135,8 +135,24 @@
       - app
       - common
       - exceptions
+      - plugin
```
seandst: the actual fix for platform virtualenv weirdness

ansible/roles/dev/tasks/main.yml
```diff
@@ -135,8 +135,24 @@
...
+- name: Install Plugin packages into Platform Virtualenv
+  command: "{{ pulp_venv_dir }}/pulp/bin/python setup.py develop"
+  args:
+    chdir: "{{ pulp_devel_dir }}/{{ item }}"
+  # If we stick with running setup.py develop for all plugins, we'd want to do it for all available
+  # plugins. For now, it only works with pulp_file
+  # with_items: "{{ pulp_available_plugins }}"
+  with_items:
+      - pulp_file
+  when: "{{ [item] | issubset(pulp_available_plugins) }}"
```
seandst: ...but while we're messing around here, let's get pulp_file installed so you can start messing with content right away in a platform runserver!

ansible/roles/dev/tasks/main.yml
```diff
@@ -135,8 +135,24 @@
...
       - exceptions
+      - plugin
       - tasking
 
+# We don't currently have a common method for installing plugins for development. In Pulp 2, it was
+# "run pulp-dev.py -I", so maybe that's what we'll do for pulp 3 as well. For now, since only the
+# file plugin works with "python setup.py develop", we'll just try that
+# exists, installing plugins to the platform virtualenv
+- name: Install Plugin packages into Platform Virtualenv
+  command: "{{ pulp_venv_dir }}/pulp/bin/python setup.py develop"
```
asmacdo: I'm curious why the other Pulp packages need to be installed system wide, but pulp_file only needs to be in a virtualenv.

seandst: The pulp packages don't have to be installed system-wide, that's just a timesaving step to effectively install them into every virtualenv, so we don't have to loop over every virtualenv and then loop over all the pulp.* packages and install them.

Since the plugins can't even be imported, let along function, without platform, and since platform is effectively nonfunctional if no plugins are installed, it (sorta?) makes sense to get all the plugins installed there, which still allowing for working on a plugin in isolation by "workon"-ing that plugin's virtualenv.

Though, comically, the pulp_file package isn't yet installed into the pulp_file virtualenv. Probably it should be, and I can add that in here as well if you think it's worth it.

seandst:
> Though, comically, the pulp_file package isn't yet installed into the pulp_file virtualenv. Probably it should be, and I can add that in here as well if you think it's worth it.

I took a stab at this and it didn't work as expect, so I'm spinning this off into a separate task tracked in redmine.

PR #50, opened 2017-04-27
Fix replacing python-lxml if not on rpm_dependency_list

mdellweg:
Provisioning fails, if python-lxml is not in rpm_dependency_list.
This version replaces python-lxml silently and does nothing if it is not found.
If you dislike list comprehensions, i can provide another version.

### DIFF
--- a/ansible/library/pulp_facts.py
+++ b/ansible/library/pulp_facts.py
@@ -52,8 +52,9 @@ rpm_dependency_list = list(set(rpm_dependency_list))
 
 # Replace python-lxml with python2-lxml
 # https://www.redhat.com/archives/pulp-dev/2016-December/msg00042.html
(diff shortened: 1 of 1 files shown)


dralley:
Hello @mdellweg, thank you for this patch!  I don't have an issue with list comprehensions but I don't believe one is actually necessary in this case.  Perhaps just

```
if 'python-lxml' in rpm_dependency_list:
    rpm_dependency_list.remove('python-lxml')
    rpm_dependency_list.append('python2-lxml')
```

would suffice?

Thanks again 👍

mdellweg:
I'm fine with that, too. Take whichever version you like more.

dralley:
I think that version would be preferable, if you could update the PR

dralley:
👍

PR #54, opened 2017-05-19
translate new pulp_deb dependencies, to pip dependencies

mdellweg:
This is a hack, to provide the debian specific python packages,
that are required by pulp_deb and not part of official repositories,
with pip.


asmacdo:
@mdellweg, thanks for submitting this! The blacklisting part was included in this PR which was just merged:  https://github.com/pulp/devel/pull/57

Rather than install with pip here, can you add `install_requires` to the setup.py of pulp_deb?

mdellweg:
The blacklisting seems to be sufficient.
Ack for #57 
Closing this.

PR #56, opened 2017-05-24
Separate roles, flatten variables

asmacdo:
This work is intended to make our development environment more flexible, simpler to debug, and less coupled to specific tools. Specifically, the dev-playbook should work correctly with and without Vagrant.


asmacdo:
Docs to go with this change: https://github.com/pulp/pulp/pull/3051

ansible/roles/vagrant/files/motd
```diff
@@ -1,31 +1,11 @@
...
-
-  Run 'phelp' to list the available functions and what they do.
-
-* phttp is an alias that gives you authenticated access to the rest API using httpie:
-
-    phttp https://localhost/pulp/api/v2/repositories/zoo/
-
-* You can ssh into your vagrant environment with vagrant ssh, but presumably
-  you already know this since you are reading this message :)
+* The 'phelp' alias will list useful functions and what they do.
```
werwty: This message displays even if phelp isn't aliased. 
When I vagrant up the minimal-playbook `supplement_bashrc` is false and the alias.bashrc is not copied over.

ansible/roles/vagrant/files/motd
```diff
@@ -1,31 +1,11 @@
...
 
 More information can be found at
 https://pulp.readthedocs.org/en/latest/dev-guide/ and you can ask questions in
```
werwty: Can you change the link to the 3.0 docs?

ansible/roles/dev/files/django.bashrc
```diff
@@ -0,0 +1 @@
+export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
```
werwty: Maybe set this as an environment variable inside the common/tasks/main.yml
http://docs.ansible.com/ansible/faq.html#how-can-i-set-the-path-or-any-other-environment-variable-for-a-task-or-entire-playbook

ansible/example-playbook.yml
```diff
@@ -0,0 +1,23 @@
+---
```
jortel: 👍 renamed to example.

ansible/minimal-playbook.yml
```diff
@@ -0,0 +1,10 @@
+---
+- hosts: dev
+  become: true
+  become_method: sudo
+  roles:
+    - vagrant
```
jortel: I'm surprised to see _vagrant_ in the minimal playbook.

ansible/roles/dev/files/django.bashrc
asmacdo: I am pretty sure that the `environment` keyword only works within Ansible scope. If we want to use Django from the shell this variable is not required because the user can also pass their own location.

PR #60, opened 2017-06-20
Add netrc file to store admin credentials

daviddavis:
This will allow us to make calls via httpie without having to supply credentials every time. More info:

https://github.com/jakubroztocil/httpie#netrc

Before:

```
http --auth admin:admin :3000/api/v3/repositories/
```

After:

```
http :3000/api/v3/repositories/
```

Note that you can still override credentials by setting `--auth`.

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -58,3 +58,10 @@
     dest: "{{ unprivileged_homedir }}/.vimrc"
     force: no
   become: false
+
+- name: Add a basic .netrc file if not present
+  copy:
+    src: files/netrc
+    dest: "{{ unprivileged_homedir }}/.netrc"
+    force: no
+  become: false
--- a/ansible/roles/common/templates/alias.bashrc
+++ b/ansible/roles/common/templates/alias.bashrc
@@ -140,4 +140,4 @@ phelp() {
 }
 _phelp_help="Print this help"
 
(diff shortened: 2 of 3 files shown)


ansible/roles/common/templates/alias.bashrc
```diff
@@ -140,4 +140,4 @@ phelp() {
 }
 _phelp_help="Print this help"
 
-alias phttp="http --verify no --cert {{ unprivileged_homedir }}/.pulp/user-cert.pem"
+alias phttp="http"
```
daviddavis: This didn't work before and has become useless but I figured I'd keep it around in case people were still using it.

ansible/roles/common/tasks/main.yml
```diff
@@ -58,3 +58,10 @@
     dest: "{{ unprivileged_homedir }}/.vimrc"
     force: no
   become: false
+
+- name: Add a basic .netrc file if not present
+  copy:
+    src: files/netrc
+    dest: "{{ unprivileged_homedir }}/.netrc"
+    force: no
+  become: false
```
bmbouter: I suspect this is not needed because I think it defaults to `false`, but keeping it here now is fine for consistency.

asmacdo: `become` is set to true in the playbook

ansible/roles/common/templates/alias.bashrc
asmacdo: I say remove it. We don't need to stay backwards compatible

PR #61, opened 2017-06-20
Support Python 3.4 in checkout.py

daviddavis:
I'm on centos 7 and the latest version of python in epel7 is python 3.4 (see https://dl.fedoraproject.org/pub/epel/7/x86_64/p/). The `subprocess.run()` is python 3.5+ only. 

I'm happy to close this out though if we decide we don't want to support python 3.4.


daviddavis:
ping @bmbouter

PR #78, opened 2017-08-31
Rewrite function for finding Pulp source code dirs

Ichimonji10:
Don't create a variable with the same name as its enclosing function.
There's no good reason to shadow names like this.

Rewrite the comments at the head of the function, and turn them into a
docstring.

Rewrite the logic that asserts that the source code for the Pulp
platform is present.

### DIFF
--- a/ansible/library/pulp_facts.py
+++ b/ansible/library/pulp_facts.py
@@ -50,27 +50,27 @@ def selinux_enforcing():
 
 
 def available_repositories(module):
-    # determine which of the platform and plugin are available,
-    # failing if platform is not available
-    available_repositories = []
+    """Find the source code for the Pulp platform and its plugins.
 
+    Raise an exception if the source code for the platform is absent.
+    """
+    _available_repositories = []
     try:
         repos = os.listdir(module.params['devel_dir'])
     except FileNotFoundError:
         repos = []
-
-    for dirname in repos:
-        if os.path.isdir(os.path.join(module.params['devel_dir'], dirname)):
-            if dirname in repositories:
-                available_repositories.append(dirname)
-
-        # if the dirname is pulp, and it doesn't pass the isdir check, there's no
-        # platform repository in the expected location. Time to explode.
-        elif dirname == 'pulp':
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against an F25 VM. When this check fails, it spits out a message like:

```
TASK [pulp-user : Gather Pulp facts] ********************************************************************************************************************************************************************************
fatal: [fedora-25-test]: FAILED! => {"changed": false, "failed": true, "msg": "The source code for the Pulp platform must be present. However, /home/vagrant/devel/pulp is either absent or is not a directory. Abort
ing.", "name": "pulp"}
```

ansible/library/pulp_facts.py
```diff
@@ -50,27 +50,26 @@ def selinux_enforcing():
...
-            msg = ('pulp repo not found in {} on remote machine, '
-                   'unable to proceed'.format(module.params['devel_dir']))
-            module.fail_json(name='pulp', msg=msg)
-    return available_repositories
+    for repo in repos:
+        if (os.path.isdir(os.path.join(module.params['devel_dir'], repo)) and
+                repo in repositories):
+            available_repositories_.append(repo)
+    if 'pulp' not in repos:
+        module.fail_json(name='pulp', msg=(
```
dralley: I suppose this is a subjective thing, but I find the code much cleaner when it's separated like it was previously.

ansible/library/pulp_facts.py
```diff
@@ -50,27 +50,26 @@ def selinux_enforcing():
...
 
 def available_repositories(module):
-    # determine which of the platform and plugin are available,
-    # failing if platform is not available
-    available_repositories = []
+    """Find the source code for the Pulp platform and its plugins.
 
+    Raise an exception if the source code for the platform is absent.
+    """
+    available_repositories_ = []
```
dralley: It would be better to name this ```available_repos``` or ```_available_repositories```.  The trailing underscore is easy to overlook.

Ichimonji10: Sure. My intent is to ensure that this variable name doesn't shadow `available_repos()`.

PR #80, opened 2017-08-31
Rewrite user creation roles

Ichimonji10:
The "common" role depends on the "vagrant" role. However, this
dependency isn't explicit. This is problematic, because a playbook
writer could incorrectly list just the "common" role in a playbook. This
can be fixed by doing one of the following:

* Define the dependency in `roles/common/meta/main.yml`.
* Merge the two roles.

Do the latter, because:

* I can think of no good use case for calling just one of the roles.
* The two roles perform a single logically coherent task. They both
  prepare a user's home directory for a Pulp installation.
* It gets us closer to a situation where there can be one role per Pulp
  component installed. This code might eventually be turned in to a role
  named "pulpcore."

The current collection of roles assume that Pulp is being installed on a
Vagrant VM, that the connection to be VM is being made as user
"vagrant," and that a development environment should be set up. It would
be nice if they instead assumed that Pulp is being installed on *some*
officially supported distribution (e.g. Fedora 26 & RHEL 7), that the
connection to the VM is made as *some* user with sudo privileges, and
that a development environment should optionally be set up. Move in this
direction by replacing the "vagrant" and "common" roles with a
"pulp-user" role that takes these new goals in to account.

Notably, many variables are now defined by the "pulp-user" role, instead
of being defined in the `host_vars/all` file. This is necessary because
many variables are derived from a value that can't be known until
runtime: the path to the target user's home directory. As a nice
side-effect, all user-defined variables now have a clear owner, in the
form of some role or another. (Where a "user-defined" variable is in
opposition to a special variable provided by Ansible.) In addition, the
variables owned by the "pulp-user" role are now documented in its
readme.

This change is not intended to change how developers use Vagrant during
their development of Pulp 3. But beware that some changes, like variable
renames, inevitably do impact users.


Ichimonji10:
This PR starts pursuing a longer-term vision where we can have a series of Ansible roles, like:

* pulpcore
* pulp-file
* pulp-ostree
* pulp-puppet

...where each of those roles is capable of installing Pulp either in a bare-bones fashion (good for QE or real-world deployments), or in a developer-friendly fashion. And we could implement more install strategies too.

This PR doesn't move us radically far in that direction. Not at all. What this PR *does* do is help make these roles useful for environments outside of Vagrant. The user creation code no longer has hard-coded references to Vagrant. Instead, one specifies a user name ("vagrant," "pulp," etc), and the roles make a suitable user by that name. This necessitates some other changes, like removing the hard-coded variables from `group_vars/all`, and instead letting the new `pulp-user` role define them at runtime, where they will be visible until the end of the playbook run. If you want to get a good sense for what this playbook does, the single best thing you can do is read `roles/pulp-user/README.md`.

I've tested these changes against almost-vanilla F25 VMs on libvirt. **A Vagrant user should test this PR.**

Ichimonji10:
Updated. The applied diff is as follows:

```diff
diff --git a/ansible/pulp-from-source.yml b/ansible/pulp-from-source.yml
index 2d395e1..db00eda 100644
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -9,6 +9,12 @@
   become_method: sudo
   vars:
     pulp_user: vagrant
+  pre_tasks:
+    # Developers want to know if Pulp works with the latest packages.
+    - name: Upgrade all currently installed packages
+      package:
+        name: '*'
+        state: latest
   roles:
     - pulp-user
     - db
diff --git a/ansible/roles/dev_tools/tasks/main.yml b/ansible/roles/dev_tools/tasks/main.yml
index 5bb4118..1fe1e97 100644
--- a/ansible/roles/dev_tools/tasks/main.yml
+++ b/ansible/roles/dev_tools/tasks/main.yml
@@ -1,3 +1,13 @@
+# This role could be made more idempotent by requesting that packages be
+# "present," not "latest." The vim packaging issue could be solved with logic
+# that says "if the current platform is Fedora and vim-enhanced isn't installed,
+# let vim-minimal be latest."
+
+# Fedora has a packaging problem where vim-enhanced will fail to install if
+# vim-minimal is out of date.
+- name: Upgrade vim-minimal
+  package: name=vim-minimal state=latest
+
 - name: Install optional useful packages
   package: name={{ item }} state=latest
   with_items:
```

Ichimonji10:
Thanks for testing. :smile:

I don't have commit privileges, so if someone else will need to press the "merge" button.

PR #81, opened 2017-09-05
Use a single variable to control dev extras

Ichimonji10:
Drop the `pip_install_extras` and `supplement_bashrc` variables. In
their place, add a variable named `pulp_install_strategy`:

> The strategy to use when creating and configuring this user, and when
> installing Pulp and its plugins. If "barebones," perform the minimum
> tasks needed to create a functional Pulp installation. If
> "development," install code from source in editable mode, install
> extra dependencies, install Bash aliases, etc. If some other value,
> raise an error.

Why do this? Well, consider what might happen if we want to support the
use case of installing Pulp in a different manner — say, from source,
but not in editable mode or with developer tools. How would this be
done?

With the current strategy, we might need to add another variable called
`install_from_source` or `pip_install_source` or somesuch. While this
might work, this produces an exponential increase in the number of
variable combinations. With n variables, there are 2^n variable
combinations. (Or more, if a variable isn't a boolean.) And presumably,
many of those combinations wouldn't be valid.

By using a single variable to control how Pulp is installed, the number
of variable combinations scales linearly. If there are three values for
the `pulp_install_strategy` variable, then there are three variable
values to test out. Hopefully, this is also more straightforward and
easy to understand.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -1,4 +1,1 @@
 ansible_python_interpreter: '/usr/bin/python3'
-# If enabled, `phelp` and other useful aliases are sourced by your bashrc.
-# In adition, some roles will add other relevant bashrc supplements
-supplement_bashrc: true
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -7,6 +7,7 @@
 - hosts: pulp3_dev
   become: true
   vars:
+    pulp_install_strategy: development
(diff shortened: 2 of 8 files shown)


Ichimonji10:
I don't think this is the right time to merge something like this. #82 needs to be considered first.

I think that there's a huge amount of value in having developers and end users all work from the same set of Ansible code, and in avoiding duplication of code. I hope we can accomplish this in the future.

PR #82, opened 2017-09-06
Publish ansible roles to install pulp3

ipanova:
closes #2840

1. pulp3, pulp3_db, pulp3_systemd will be eventually published to the galaxy
2. pulp3_migrations role still needs to be written and published to the galaxy
3. rabbitmq and qpidd roles from jtyr needs to be adopted for python3 (AI: reach jtyr)
4. the docs from readme will be moved as a separate PR to pulp/pulp installation section.

### DIFF
--- a/deploy_pulp3/README.md
+++ b/deploy_pulp3/README.md
@@ -1,0 +1,6 @@
+1. install ansible2.2+, python3, virtualenv, libselinux-python
+2. download roles from galaxy
+   ansible-galaxy install -r requirements.yml -p ./roles
+3. run the playbook by specifying which broker you want rabbitmq|qpidd
+   ansible-playbook deploy-pulp3.yml --extra-vars "pulp3_broker=rabbitmq"
+
--- a/deploy_pulp3/deploy-pulp3.yml
+++ b/deploy_pulp3/deploy-pulp3.yml
@@ -1,0 +1,20 @@
+---
+- hosts: all
+  become: true
+  vars:
+    ansible_python_interpreter: "/usr/bin/python3"
+  pre_tasks:
+    - name: Require minimal Python and Ansible versions
+      assert:
+        that:
+          - "ansible_python_version|version_compare('3.5.0', operator='ge')"
+          - "ansible_version.full|version_compare('2.2.0', operator='ge')"
+        msg: >
+          This role and those that depend on it require at least Python 3.5 and
+          Ansible 2.2.
+  roles:
+    - pulp3
+    - pulp3_db
+    - pulp3_systemd
+    - { role: rabbitmq, when: pulp3_broker == "rabbitmq"}
+    - { role: qpid_cpp_server, when: pulp3_broker == "qpidd"}
--- a/deploy_pulp3/pulp3/defaults/main.yml
+++ b/deploy_pulp3/pulp3/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+pulp_user: pulp
--- a/deploy_pulp3/pulp3/tasks/main.yml
+++ b/deploy_pulp3/pulp3/tasks/main.yml
@@ -1,0 +1,107 @@
+---
+- name: Create user {{ pulp_user }}
+  user:
+    name: "{{ pulp_user }}"
+    comment: "Pulp user"
+
+- name: Get user's info
+  getent:
+    database: passwd
+    key: "{{ pulp_user }}"
+
+- name: Set variable referencing user's home directory
+  set_fact:
+    pulp_user_home: "{{ getent_passwd[pulp_user][4] }}"
+
+- name: Make user's home directory world-readable
+  file:
+    path: "{{ pulp_user_home }}"
+    state: directory
+    mode: 0755
+
+- name: Set variable referencing pulpvenv dir
+  set_fact:
+    pulp_venv: "{{ pulp_user_home }}/pulpvenv"
+
+- name: Create ~/.bashrc.d/
+  file:
+    path: "{{ pulp_user_home }}/.bashrc.d/"
+    state: directory
+  become_user: "{{ pulp_user }}"
+
+- name: Create directory /var/lib/pulp
+  file:
+    path: /var/lib/pulp
+    state: directory
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+    mode: 0775
+
+- name: Install pulpcore
+  pip:
+    name: pulpcore
+    state: present
+    virtualenv: "{{ pulp_venv }}"
+    virtualenv_python: python3
+  become_user: "{{ pulp_user }}"
+
+- name: Add Django supplemental bashrc
+  copy:
+    src: files/django.bashrc
+    dest: "{{pulp_user_home}}/.bashrc.d/django.bashrc"
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+
+- name: Source  ~/.bashrc.d/
+  blockinfile:
+    path: "{{ pulp_user_home }}/.bashrc"
+    block: |
+      if [ -d ~/.bashrc.d ]; then
+        for file in ~/.bashrc.d/*; do
+          . "$file"
+        done
+      fi
+    marker: "# {mark} Ansible managed block: source bashrc.d/"
+    create: yes
+  become_user: "{{ pulp_user }}"
+
+- name: Create /etc/pulp dir
+  file:
+    path: /etc/pulp
+    state: directory
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+
+- name : Find server.yaml config file
+  find:
+    paths: "{{ pulp_venv }}"
+    patterns: 'server.yaml'
+    recurse: true
+  register: result
+
+- name: Ensure the server.yaml was found
+  assert:
+    that: 'result["matched"] == 1'
+
+- name: Install server.yaml Config file
+  copy:
+    remote_src: true
+    src: '{{ result["files"][0]["path"] }}'
+    dest: "/etc/pulp/server.yaml"
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+    force: false
+
+- name: Generate random secret for Django
+  shell: cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(\-_=+)' | head -c 50
+  check_mode: false
+  changed_when: false
+  register: random_secret
+
+- name: Add development server configuration in server.yaml
+  blockinfile:
+    path: "/etc/pulp/server.yaml"
+    block: |
+      # Pulp Development Configuration
+      DEBUG: True
+      SECRET_KEY: {{ random_secret.stdout | to_nice_yaml }}
--- a/deploy_pulp3/pulp3_db/files/pg_hba.conf
+++ b/deploy_pulp3/pulp3_db/files/pg_hba.conf
@@ -1,0 +1,14 @@
+# PostgreSQL Client Authentication Configuration File
+# ===================================================
+#
+# Refer to the "Client Authentication" section in the PostgreSQL
+# documentation for a complete description of this file.
+
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     peer
+# IPv4 local connections:
+host    all             all             127.0.0.1/32            trust
+# IPv6 local connections:
+host    all             all             ::1/128                 trust
--- a/deploy_pulp3/pulp3_db/tasks/main.yml
+++ b/deploy_pulp3/pulp3_db/tasks/main.yml
@@ -1,0 +1,68 @@
+- name: Install postgres packages
+  package: name={{ item }} state=present
+  with_items:
+      - postgresql
+      - postgresql-server
+      # needed by postgres-related tasks and other things
+      # using python outside of a virtualenv
+      - python3-psycopg2
+
+- name: Set up postgres data dir
+  command: postgresql-setup --initdb
+  args:
+      creates: /var/lib/pgsql/data/base
+
+- name: Set up postgress access rules
+  copy:
+      src: pg_hba.conf
+      dest: /var/lib/pgsql/data/pg_hba.conf
+      owner: postgres
+      group: postgres
+      mode: 0600
+
+- name: Start and enable postgresql
+  service: name=postgresql state=started enabled=yes
+
+- name: Set up pulp DB user
+  postgresql_user:
+      name: "{{ pulp_user }}"
+      role_attr_flags: SUPERUSER,LOGIN
+  become: true
+  become_user: postgres
+
+- name: Create pulp database
+  postgresql_db:
+      name: "{{ pulp_user }}"
+      owner: "{{ pulp_user }}"
+  become: true
+  become_user: postgres
+
+- name : Find manage.py
+  find:
+    paths: "{{ pulp_venv }}"
+    patterns: 'manage.py'
+    recurse: true
+  register: result
+
+- name: Ensure the manage.py was found
+  assert:
+    that: 'result["matched"] == 1'
+
+- name: Changing perm of manage.py, adding "+x"
+  file: dest={{ result["files"][0]["path"] }} mode=a+x
+
+- name: Run Django migrations
+  django_manage:
+    command: "{{ item }}"
+    app_path: '{{ result["files"][0]["path"]| dirname}}'
+    virtualenv: "{{ pulp_user_home }}/pulpvenv"
+  with_items:
+    - 'makemigrations pulp_app'
+    - 'migrate --noinput auth'
+    - 'migrate --noinput'
+    - 'reset-admin-password --password admin'
+  become: true
+  become_user: "{{ pulp_user }}"
+
+- name: Changing perm of manage.py, adding "-x"
+  file: dest={{ result["files"][0]["path"] }} mode=a-x
--- a/deploy_pulp3/pulp3_systemd/tasks/main.yml
+++ b/deploy_pulp3/pulp3_systemd/tasks/main.yml
@@ -1,0 +1,30 @@
+- name: Install Systemd unit files
+  # The template allows us to specify the platform python environment.
+  template:
+      src: "templates/{{ item }}.j2"
+      dest: "/etc/systemd/system/{{ item }}.service"
+  with_items:
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_worker@
+  register: result
+ 
+- name: Create /var/cache/pulp directory
+  file:
+    path: /var/cache/pulp
+    state: directory
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+
+- name: reload systemd units
+  systemd:
+    daemon-reload: yes
+    name: "{{ item }}"
+    state: restarted
+    enabled: yes
+  with_items:
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_worker@1
+    - pulp_worker@2
+  when: result|changed
--- a/deploy_pulp3/pulp3_systemd/templates/pulp_celerybeat.j2
+++ b/deploy_pulp3/pulp3_systemd/templates/pulp_celerybeat.j2
@@ -1,0 +1,14 @@
+[Unit]
+Description=Pulp's Celerybeat
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/default/pulp_celerybeat
+User={{ pulp_user }}
+WorkingDirectory=/var/run/pulp_celerybeat/
+RuntimeDirectory=pulp_celerybeat
+ExecStart={{ pulp_venv }}/bin/celery beat --app=pulpcore.tasking.celery_app:celery --scheduler=pulpcore.tasking.services.scheduler.Scheduler
+
+[Install]
+WantedBy=multi-user.target
--- a/deploy_pulp3/pulp3/files/django.bashrc
+++ b/deploy_pulp3/pulp3/files/django.bashrc
@@ -1,0 +1,1 @@
+export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
--- a/deploy_pulp3/pulp3_systemd/templates/pulp_resource_manager.j2
+++ b/deploy_pulp3/pulp3_systemd/templates/pulp_resource_manager.j2
@@ -1,0 +1,16 @@
+[Unit]
+Description=Pulp Resource Manager
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/default/pulp_resource_manager
+User={{ pulp_user }}
+WorkingDirectory=/var/run/pulp_resource_manager/
+RuntimeDirectory=pulp_resource_manager
+ExecStart={{ pulp_venv }}/bin/celery worker -A pulpcore.tasking.celery_app:celery -n resource_manager@%%h\
+          -Q resource_manager -c 1 --events --umask 18\
+          --pidfile=/var/run/pulp_resource_manager/resource_manager.pid
+
+[Install]
+WantedBy=multi-user.target
--- a/deploy_pulp3/pulp3_systemd/templates/pulp_worker@.j2
+++ b/deploy_pulp3/pulp3_systemd/templates/pulp_worker@.j2
@@ -1,0 +1,17 @@
+[Unit]
+Description=Pulp Celery Worker
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/default/pulp_workers
+EnvironmentFile=-/etc/default/pulp_workers_%i
+User={{ pulp_user }}
+WorkingDirectory=/var/run/pulp_worker_%i/
+RuntimeDirectory=pulp_worker_%i
+ExecStart={{ pulp_venv }}/bin/celery worker -A pulpcore.tasking.celery_app:celery\
+          -n reserved_resource_worker_%i@%%h -c 1 --events --umask 18\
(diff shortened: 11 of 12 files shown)


deploy_pulp3/pulp3_systemd/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+    - pulp_resource_manager
+    - pulp_worker@
+- name: Create /var/cache/pulp directory
+  file:
+    path: /var/cache/pulp
+    state: directory
+    owner: "{{pulp_user}}"
+    group: "{{pulp_user}}"
+
+- name: reload systemd units
```
ammaritiz: The `systemd` module now has a `daemon_reload` attribute which could be used instead.
[http://docs.ansible.com/ansible/latest/systemd_module.html](url)

deploy_pulp3/pulp3_systemd/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  command: systemctl daemon-reload
```
ammaritiz: The `systemd` module now has a `daemon_reload` attribute which could be used instead.
[http://docs.ansible.com/ansible/latest/systemd_module.html](url)

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
+---
+# tasks file for pulp3
+- name: Create user {{ pulp_user }}
+  user:
+    name: "{{ pulp_user }}"
+    comment: "Pulp user"
+
+- name: Get user's home directory
+  shell: 'getent passwd "{{ pulp_user }}" | cut -d: -f6'
```
ammaritiz: There is a [getent](http://docs.ansible.com/ansible/latest/getent_module.html) module which could achieve this in a much more readable manner, something like this:
```
- name: Get user's home directory
    getent:
     database: passwd
     key: "{{ pulp_user }}"
     split: ':'
- debug:
   var: getent_passwd['{{ pulp_user }}'][4]
```

deploy_pulp3/pulp3_systemd/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+# defaults file for pulp3
+pulp_user: pulp
+pulp_venv : "/home/pulp/pulpvenv"
```
Ichimonji10: How about setting this at runtime with the `getent` and `set_fact` modules? The Pulp user's home directory could be in any number of places.

deploy_pulp3/pulp3_systemd/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+# defaults file for pulp3
```
Ichimonji10: Redundant comment is redundant?

deploy_pulp3/pulp3_db/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+  package: name={{ item }} state=present
+  with_items:
+      - postgresql
+      - postgresql-server
+      # needed by postgres-related tasks and other things
+      # using python outside of a virtualenv
+      - python3-psycopg2
+
+- name: Set up postgres data dir
+  command: postgresql-setup --initdb
```
Ichimonji10: The `command` module typically does things unconditionally, which makes plays that use it non-idempotent. Could this be wrapped in a conditional?

```yaml
- name: Install postgresql
  ...
  register: result

- name: Initialize postgresql's database
  ...
  when: result|changed
```

deploy_pulp3/pulp3_db/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: Set up postgress access rules
+  copy:
+      src: pg_hba.conf
+      dest: /var/lib/pgsql/data/pg_hba.conf
+      owner: postgres
+      group: postgres
+      mode: 0600
+
+- name: Start and enable postgresql
+  service: name=postgresql state=restarted enabled=yes
```
Ichimonji10: Same question about idempotency. Could this be wrapped in a conditional?

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+  pip:
+    name: pulpcore
+    state: latest
+    virtualenv: "{{ pulp_user_home }}/pulpvenv"
+    virtualenv_python: python3
+  become_user: "{{ pulp_user }}"
+
+- name: Add Django and Pulp supplemental bashrc
+  copy:
+    src: files/supplem.bashrc
```
ammaritiz: Even if you just specify the filename, Ansible will pick it up automatically from the `files` directory of the role.

Ichimonji10:
Much of this code is a duplicate of existing Ansible code used by develoeprs. And as this expands, the duplication will likely increase. That means we end up creating *more* work for ourselves, where there's both user-facing and developer-facing code that both need to be maintained. I don't think we want two sets of Ansible code to maintain.

Instead, could we take our existing Ansible code that the developers use and re-work it to allow users to install Pulp either from source or from PyPI, and eventually re-work that into discrete roles? I think it's doable.

It's true that this would mean initially telling end users that "you can install Pulp 3 by cloning this repository, checking out this branch, and then running this playbook." But we're not even in alpha, and given the benefits listed - much less duplication, ad an evolutionary instead of big-bang design approach - wouldn't it be worth it?

deploy_pulp3/pulp3/tasks/main.yml
Ichimonji10: I find it very useful to always specify files relative to the `files` directory. That lets me write code like this:

```yaml
- name: Install systemd units
  copy:
    src: etc_systemd_system/
    dest: /etc/systemd/system/
    owner: root
    group: root
    mode: 0644
```

In this example, all the files in `files/etc_systemd_system/` are copied into the target directory.

deploy_pulp3/deploy-pulp3.yml
```diff
@@ -0,0 +1,19 @@
+---
+- hosts: localhost
+  become: true
```
asmacdo: vars:
        ansible_python_interpreter = "/usr/bin/python3"

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+    create: yes
+  become_user: "{{ pulp_user }}"
+
+- name: Create /etc/pulp dir
+  file:
+    path: /etc/pulp
+    state: directory
+
+- name: Install server.yaml Config file
+  copy:
```
Ichimonji10: I think `force: false` would be a good addition to this call. When this play runs, the following will happen:

1. `server.yaml` will be copied from the venv to `/etc/pulp/server.yaml`.
2. The `DEBUG: True` and `SECRET_KEY: ...` lines will be added.

If this play is run again, both of these steps will run. Step 1 will run because the source and destination files differ, and step 2 will run because the requested block is missing. The result is a non-idempotent play. Using `force: false` sidesteps this issue.

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+    remote_src: true
+    src: "{{pulp_venv}}/lib/python3.5/site-packages/pulpcore/etc/pulp/server.yaml"
+    dest: "/etc/pulp/server.yaml"
+
+- name: Generate random secret for Django
+  shell: cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(\-_=+)' | head -c 50
```
Ichimonji10: The following could be a good addition to this task:

```yaml
check_mode: false
changed_when: false
```

`changed_when: false` tells Ansible that this task doesn't change anything on the remote system even when it runs, thus avoiding spurious "changed" items in Ansible's final report.

`check_mode: false` lets this task run even when `--check` is passed, meaning that `random_secret` will be set even when `--check` is set, which lets the subsequent task be checked without spurious errors.

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+
+- name: Create /etc/pulp dir
+  file:
+    path: /etc/pulp
+    state: directory
+
+- name: Install server.yaml Config file
+  copy:
+    remote_src: true
+    src: "{{pulp_venv}}/lib/python3.5/site-packages/pulpcore/etc/pulp/server.yaml"
```
Ichimonji10: Rather than hard-coding this path, one could use the `find` module to find `server.yaml`. This will help when supporting multiple versions of Python (say, 3.5 and 3.6).

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+  shell: cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(\-_=+)' | head -c 50
+  register: random_secret
+
+- name: Add development server configuration in server.yaml
+  blockinfile:
+    path: "/etc/pulp/server.yaml"
+    block: |
+      # Pulp Development Configuration
+      DEBUG: True
+      SECRET_KEY: {{ random_secret.stdout | to_nice_yaml }}
```
Ichimonji10: It might be better to use two `lineinfile` tasks instead of this `blockinfile` task. Again, this is to promote idempotency. Given that `random_secret` will vary every time this play runs, this block will be updated every time this task runs, which is not Ansible-ish.

deploy_pulp3/pulp3_systemd/templates/pulp_celerybeat.j2
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Pulp's Celerybeat
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/default/pulp_celerybeat
+User=apache
```
Ichimonji10: This user probably shouldn't be hard-coded.

deploy_pulp3/README.md
```diff
@@ -0,0 +1,6 @@
+1. install ansible2.2+, python3, virtualenv, libselinux-python
+2. download roles from galaxy
+   ansible-galaxy install -r requirements.yml -p ./roles
+3. run the playbook by specifying which broker you want rabbitmq|qpidd
+   ansible-playbook deploy-pulp3.yml --extra-vars "pulp3_broker=rabbitmq"
+
```
Ichimonji10: It might be worth mentioning that `deploy-pulp3.yml` can be used as a sample playbook.

deploy_pulp3/pulp3/tasks/main.yml
werwty: We should not be setting debug: True and generating a secret key for a non dev install of pulp. This task and the previous task should be removed.

deploy_pulp3/pulp3_db/tasks/main.yml
ipanova: @Ichimonji10 i think we can safely leave as it is because this step will not be executed if the file already exists.

deploy_pulp3/pulp3_db/tasks/main.yml
ipanova: honestly saying i do not see much or harm in doing the restart here, i do not know but i guess if there is a change in the pg_hba.conf then we need to restart the service, that means it  forces me to check not just the first task but also the second

deploy_pulp3/pulp3_systemd/tasks/main.yml
ipanova: i guess that works just with ansible 2.4, in previous version i'd need to provide the name of the service, in thise case there are 3 of them so would it be shorter to leave this as it is?

deploy_pulp3/pulp3_db/tasks/main.yml
Ichimonji10: Sure. This is safe. But as written, Ansible will always report this step as having changed something, even if this command doesn't actually change anything.

ipanova: it shows to me as 'ok' and not 'changed'

deploy_pulp3/pulp3_db/tasks/main.yml
Ichimonji10: Is this restart harmful? No. But again, idempotency is a desirable property of Ansible playbooks.

Handlers might be an appropriate solution to the problem of needing to check two tasks:

> These ‘notify’ actions are triggered at the end of each block of tasks in a play, and will only be triggered once even if notified by multiple different tasks.

deploy_pulp3/pulp3/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+pulp_user: pulp
+pulp_venv : "/home/pulp/pulpvenv"
```
Ichimonji10: If someone overrides `pulp_user`, then do they also have to override `pulp_venv`? This seems like something that can be derived with the getent module.

ipanova: as far as i understood we will stick to pulp user instead of apache.

deploy_pulp3/deploy-pulp3.yml
```diff
@@ -0,0 +1,21 @@
+---
+- hosts: localhost
```
Ichimonji10: Can this *not* be localhost? What if I want to install Pulp on, say, some other machine? Let this be `hosts: all`, and have the user specify the hosts when calling this playbook.

ipanova: yes this is how it is going to be once the 'wip' label will be removed ;)

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+    mode: 0775
+
+- name: Install pulpcore
+  pip:
+    name: pulpcore
+    state: latest
+    virtualenv: "{{ pulp_user_home }}/pulpvenv"
+    virtualenv_python: python3
```
Ichimonji10: How about specifying `virtualenv_command: python3 -m venv` instead? Doing this drops the dependency on virtualenv.

deploy_pulp3/README.md
```diff
@@ -0,0 +1,6 @@
+1. install ansible2.2+, python3, virtualenv, libselinux-python
```
Ichimonji10: Which virtualenv package should users install? python2-virtualenv? python3-virtualenv? Fedora 25 has no virtualenv package.

deploy_pulp3/pulp3/defaults/main.yml
Ichimonji10: And? If a user is allowed to change under which user Pulp is installed, let's do it well.

deploy_pulp3/deploy-pulp3.yml
Ichimonji10: Great. :+1:

deploy_pulp3/pulp3_db/files/pg_hba.conf
```diff
@@ -0,0 +1,14 @@
+# PostgreSQL Client Authentication Configuration File
+# ===================================================
+#
+# Refer to the "Client Authentication" section in the PostgreSQL
+# documentation for a complete description of this file.
+
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     trust
```
werwty: I would recommend `local pulp {{ pulp_user }} peer`  as the permissions for this and the host connections.

deploy_pulp3/pulp3_db/tasks/main.yml
werwty: A handler would not be appropriate here, since the postgresql has to be running before database and user can use it. I would suggest `state: started`

deploy_pulp3/pulp3_db/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+
+- name: Set up pulp DB user
+  postgresql_user:
+      name: pulp
```
werwty: I think we should use the psql peer authentication method, which means this should match the unix user name.

deploy_pulp3/README.md
werwty: I believe python3-virtualenv should be used for pulp3

werwty:
@ipanova since we're really not doing anything special with our database setup (install, create user, create database) it might be worth it to consider using someone else's psql ansible role. 
At a quick glance https://github.com/ANXS/postgresql does everything we need.

deploy_pulp3/README.md
Ichimonji10: Hunh. I tried that, but the playbook failed to run.

Ichimonji10:
Beware that ANXS/postgresql is a bit [complex](https://github.com/ANXS/postgresql/blob/master/defaults/main.yml) and [needs help](https://github.com/ANXS/postgresql/blob/master/README.md). Given that our current database configuration procedure is simple, the costs of adding another dependency (breakage and bugs that we don't know about and don't control) may outweigh the benefits (saving a few lines of code).

deploy_pulp3/pulp3_db/tasks/main.yml
Ichimonji10: Ah. Because of the `creates`. I either missed that or it changed at some point. Never mind.

deploy_pulp3/pulp3_db/tasks/main.yml
Ichimonji10: Of course. Handlers *alone* wouldn't be appropriate. But triggering a restart of this service with a handler may be appropriate.

deploy_pulp3/pulp3_systemd/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+    - pulp_worker@
+- name: Create /var/cache/pulp directory
+  file:
+    path: /var/cache/pulp
+    state: directory
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+
+- name: reload systemd units
+  command: systemctl daemon-reload
```
Ichimonji10: This can be dropped in favor of the `systemd` module's `daemon_reload` argument. This would also require using the `systemd` module instead of the `service` module, below.

deploy_pulp3/pulp3_db/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+      src: pg_hba.conf
+      dest: /var/lib/pgsql/data/pg_hba.conf
+      owner: postgres
+      group: postgres
+      mode: 0600
+
+- name: Start and enable postgresql
+  service: name=postgresql state=restarted enabled=yes
+
+- name: Set up pulp DB user
```
werwty: DB User and DB Create tasks should be ran as the postgresql user

werwty:
@Ichimonji10 fair enough about carrying additional dependencies. I do think that getting the database configuration right is difficult and important. We can get away with doing our own right now for an alpha  playbook, but eventually we do want something more robust.

I also want to point out that the config file you linked contains the default values, to get this working for us right now the vars we need to provide should look more like [this](https://github.com/ANXS/postgresql/blob/master/tests/vars.yml )

Ichimonji10:
Fair enough. If doing the configuration right becomes burdensome, then that's a great rationale for adding another dependency.

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,102 @@
...
+    path: /var/lib/pulp
+    state: directory
+    owner: "{{ pulp_user }}"
+    group: "{{ pulp_user }}"
+    mode: 0775
+
+- name: Install pulpcore
+  pip:
+    name: pulpcore
+    state: latest
```
Ichimonji10: Shouldn't this be "present" instead of "latest?" This role is designed for use by end users, and if an end user has a working Pulp installation, then shouldn't we leave it be instead of fiddling with packages under their feet?

Ichimonji10:
Ansible code is typically laid out with host vars in a `host_vars` directory, roles in a `roles` directory, etc. Shouldn't the roles be in a `roles` directory?

ipanova:
@Ichimonji10 Jeremy :) the roles will be published to galaxy, as i told this is just a temporary place for them here, since it is wip. as a result of my work just 2 yml files will be mergedf into this repo. requirements and the playbook itself.

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,101 @@
...
+    virtualenv: "{{ pulp_venv }}"
+    virtualenv_python: python3
+  become_user: "{{ pulp_user }}"
+
+- name: Add Django supplemental bashrc
+  copy:
+    src: files/django.bashrc
```
werwty: owner: {{ pulp_user }}

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,101 @@
...
+          . "$file"
+        done
+      fi
+    marker: "# {mark} Ansible managed block: source bashrc.d/"
+    create: yes
+  become_user: "{{ pulp_user }}"
+
+- name: Create /etc/pulp dir
+  file:
+    path: /etc/pulp
```
werwty: owner: {{ pulp_user }}

deploy_pulp3/pulp3/tasks/main.yml
```diff
@@ -0,0 +1,101 @@
...
+
+- name: Ensure the server.yaml was found
+  assert:
+    that: 'result["matched"] == 1'
+
+- name: Install server.yaml Config file
+  copy:
+    remote_src: true
+    src: '{{ result["files"][0]["path"] }}'
+    dest: "/etc/pulp/server.yaml"
```
werwty: owner: {{pulp_user}}

Ichimonji10:
OK.

deploy_pulp3/deploy-pulp3.yml
```diff
@@ -0,0 +1,21 @@
...
+  become: true
+  become_method: sudo
```
ammaritiz: Could do away with this line, since the default value for `become_method` is `sudo`.

deploy_pulp3/pulp3_systemd/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+    daemon-reload: yes
+    name: "{{ item }}"
+    state: restarted
+    enabled: yes
+  with_items:
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_worker@1
+    - pulp_worker@2
+  when: result|changed
```
ammaritiz: Since, we are using `result|changed` variable, which is the overall changed status of the 'Install Systemd unit files' task, we would end up restarting each and every service even if just one of them were installed. Shouldn't we loop over the `results` attribute inside `result`, in order to restart only that particular service? A better naming for the registered variable in that case would be something like `services_installed`.
Also, I don't understand why there is `pulp_worker@1` and `pulp_worker@2` here. Maybe, I am missing something.

Ichimonji10:
If I make sure to:

* install python2-virtualenv
* disable selinux
* edit the playbook's "hosts" line so that it doesn't reference localhost

...then I can successfully execute this playbook. After doing so, `systemctl status` reports healthy services.

That done, how is one supposed to run Pulp? In my case, I logged in to the target system and executed:

```sh
su pulp
source ~/pulpvenv/bin/activate
python3 ~/pulpvenv/lib/python3.5/site-packages/pulpcore/app/manage.py runserver 0.0.0.0:8000
```

This let me talk to Pulp 3. But is there some easier way to do this? Perhaps a systemd unit? If this problem is already solved, it would be great if the `README.md` explained how to solve the issue.

ipanova:
@werwty  I am closing this PR and as a result, here are the  repos created for roles:
https://github.com/pulp/ansible-pulp3
https://github.com/pulp/ansible-pulp3_db
https://github.com/pulp/ansible-pulp3_systemd

new PRs created:
https://github.com/pulp/pulp/pull/3148
https://github.com/pulp/devel/pull/89

ipanova:
special thanks to @Ichimonji10 for testing the playbook :star2:

Ichimonji10:
Tested against RHEL 7. It fails:

```console
$ ansible-playbook deploy-pulp3.yml -e pulp3_broker=rabbitmq -i rhel-7-test,                 

PLAY [all] ************************************************************************************************************

TASK [Gathering Facts] ************************************************************************************************
ok: [rhel-7-test]            

TASK [Require minimal Python and Ansible versions] ********************************************************************
fatal: [rhel-7-test]: FAILED! => {                         
    "assertion": "ansible_python_version|version_compare('3.5.0', operator='ge')",                                     
    "changed": false,        
    "evaluated_to": false,   
    "failed": true,          
    "msg": "This role and those that depend on it require at least Python 3.5 and Ansible 2.2.\n"                      
}                            
        to retry, use: --limit @/home/ichimonji10/code/devel/ansible/deploy-pulp3.retry                                

PLAY RECAP ************************************************************************************************************
rhel-7-test                : ok=1    changed=0    unreachable=0    failed=1
```

Python 3.5 isn't available, AFAICT.

```console
$ yum repolist
Loaded plugins: product-id, search-disabled-repos, subscription-manager
repo id                                                                                                         repo name                                                                                                                status
epel/x86_64                                                                                                     Extra Packages for Enterprise Linux 7 - x86_64                                                                           11,947
rhel-7-server-extras-rpms/x86_64                                                                                Red Hat Enterprise Linux 7 Server - Extras (RPMs)                                                                           639
rhel-7-server-optional-rpms/7Server/x86_64                                                                      Red Hat Enterprise Linux 7 Server - Optional (RPMs)                                                                      12,809
rhel-7-server-rpms/7Server/x86_64                                                                               Red Hat Enterprise Linux 7 Server (RPMs)                                                                                 17,215
repolist: 42,610
```

PR #83, opened 2017-09-06
Drop unnecessary directive from Ansible playbook

Ichimonji10:
The default value for `become_method` is "sudo." Don't re-declare this
option in the `pulp-from-source.yml` playbook. It's acts as unnecessary
noise that distracts readers from more interesting code.

### DIFF
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -6,7 +6,6 @@
 # install Bash aliases.
 - hosts: pulp3_dev
   become: true
-  become_method: sudo
   vars:
     pulp_user: vagrant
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against several Fedora 25 VMs. In one case, I logged in as root, and in another case, I logged in as a throw-away user named "goober" that belonged to group "wheel."

Ichimonji10:
We'll eventually want to drop the "become" declaration from `pulp-from-source.yml`. Leaving `become: true` in the playbook means that our roles will assume that `become` is true, and this will make our roles less portable.

Ichimonji10:
:tada:

PR #84, opened 2017-09-06
Simplify SELinux management code

Ichimonji10:
`library/pulp_facts.py` currently has a function that sets a fact named
`pulp_selinux_enforcing`. This fact tells whether SELinux is in the
"enforcing" state or not.

Don't bother doing this. It's pointless. It doesn't help the "selinux"
module do its job. It pollutes the facts namespace with an extra
variable. It's yet more code to maintain. In short, it's a liability.

Also, drop some unused imports.

### DIFF
--- a/ansible/library/pulp_facts.py
+++ b/ansible/library/pulp_facts.py
@@ -3,9 +3,6 @@
 This module gathers Pulp specific Ansible facts about the remote machine.
 """
 import os
-import platform
-import subprocess
-import tempfile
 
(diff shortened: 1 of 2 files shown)


Ichimonji10:
This needs to be reviewed by someone who uses Vagrant. I'm doing all my testing with non-Vagrant VMs.

dralley:
Tested with vagrant environment, seems to work great.

Ichimonji10:
Yay!

Rebased on upstream/3.0-dev and re-tested.

PR #85, opened 2017-09-06
Drop assertion from fact-gathering code

Ichimonji10:
When run by Ansible, `library/pulp_facts.py` exports a fact called
`pulp_available_repositories`. This fact is a list of all of the
directories in the ~/devel/ directory. When the fact-gathering code in
this module is executed, if the `pulp` directory is absent from
`pulp_available_repositories`, an exception is thrown. The idea is that
one can't install Pulp if this source code is missing.

This is problematic for two reasons.

First, this check is redundant. The `dev` role already asserts that
"pulp" is in `pulp_available_repositories`.

Second, this check makes the existing Ansible code more inflexible. What
if one wants to install Pulp not from locally available source code, but
directly from source code on GitHub, or from packages on PyPI? By
complaining about missing source code during the fact-gathering stage,
this code prevents any of those use cases from being realized.

### DIFF
--- a/ansible/library/pulp_facts.py
+++ b/ansible/library/pulp_facts.py
@@ -50,10 +50,7 @@ def selinux_enforcing():
 
 
 def available_repositories(module):
-    """Find the source code for the Pulp platform and its plugins.
-
-    Raise an exception if the source code for the platform is absent.
-    """
+    """Find the source code for the Pulp platform and its plugins."""
     _available_repositories = []
     try:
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a F25 (non-Vagrant) VM.

Ichimonji10:
Thank you!

As in other PRs, someone else will have to merge this.

Ichimonji10:
Can someone merge?

PR #86, opened 2017-09-07
Drop dependency on virtualenv executable

Ichimonji10:
Python 3.3 and newer ships with the "venv" module, which is capable of
creating and managing virtualenvs. The virtualenv executable is only
needed if using an older version of Python.

Use `python -m venv` when creating virtualenvs. This drops the
dependency on the virtualenv executable. (And whatever packages provide
that executable.) Deployments of Pulp 3 will be faster and
more reliable if they use code that ships with the standard library
instead of third-party packages.


Ichimonji10:
As usual, it would be great if a Vagrant user could test this.

Ichimonji10:
Yay!

Rebased on upstream/3.0-dev and re-tested.

PR #87, opened 2017-09-07
Make reset_db role more flexible

Ichimonji10:
The `reset_db` role executes a script called `db-reset.sh`. If Pulp is
installed from PyPI or directly from git, then it won't exist in the
Pulp user's ~/devel/pulp directory. But no matter what type of
installation, it should exist in the virtualenv into which Pulp 3 is
installed.

Update the `reset_db` role so that it can find the `db-reset.sh` script
regardless of how Pulp 3 was installed.

### DIFF
--- a/ansible/roles/reset_db/tasks/main.yml
+++ b/ansible/roles/reset_db/tasks/main.yml
@@ -1,2 +1,13 @@
-- name: Reset and Migrate Pulp DB
-  command: "{{ pulp_devel_dir }}/pulp/platform/pulpcore/app/db-reset.sh {{ pulp_venv }}"
+- name: Find db-reset.sh
+  find:
+    paths: "{{ pulp_venv }}"
+    recurse: true
+    patterns: 'db-reset.sh'
+  register: result
+
+- name: Assert one db-reset.sh file has been found
+  assert:
+    that: 'result["matched"] == 1'
+
+- name: Execute db-reset.sh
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I've tested this change with a play that installs Pulp 3 with:

```yaml
- name: Install pulpcore
  pip:
    name: git+https://github.com/pulp/pulp.git@3.0-dev#egg=pulpcore&subdirectory=platform
    virtualenv: "{{ pulp_venv }}"
    virtualenv_command: /usr/bin/python3 -m venv
  become_user: "{{ pulp_user }}"
```

ansible/roles/reset_db/tasks/main.yml
```diff
@@ -1,2 +1,13 @@
...
+- name: Find db-reset.sh
+  find:
+    paths: "{{ pulp_venv }}"
+    recurse: true
+    patterns: 'db-reset.sh'
+  register: result
+
+- name: Assert one db-reset.sh file has been found
+  assert:
+    that: 'result["files"] | length == 1'
```
ammaritiz: Minor improvement using the `matched` field:

```
- name: Assert one db-reset.sh file has been found
  assert:
   that: 'result["matched"] == 1'
```

Ichimonji10: Even better!

Ichimonji10:
Updated.

Ichimonji10:
Can someone merge?

PR #88, opened 2017-09-12
Use getent module instead of getent command

Ichimonji10:
It's usually better to use a module than an arbitrary command.


Ichimonji10:
Thanks to @ammaritiz for pointing out that the getent module exists.

PR #90, opened 2017-09-15
Revert "Make reset_db role more flexible"

asmacdo:
Reverts pulp/devel#87

The change fixed the pip/git installations, but broke "--editable" python installs (pip -e, setup.py develop). It turns out that the `*.egg-link` files created in dev installs aren't real symlinks, so this pattern won't work on dev installs.


Ichimonji10:
:+1:

dralley:
LGTM

PR #91, opened 2017-09-15
use pulp-manager to reset db

asmacdo:
https://pulp.plan.io/issues/3012


Ichimonji10:
Tested against a vanilla F25 system where:

* python2-virtualenv is installed
* SELinux is disabled

I tested this code by rebasing on upstream/3.0-dev and executing `ansible-playbook deploy-pulp3.yml -i fedora-25-test, -e pulp3_broker=rabbitmq`.

asmacdo:
To me merged with https://github.com/pulp/devel/pull/91

ansible/pulp-from-source.yml
```diff
@@ -20,8 +20,8 @@
     - broker
     - lazy
     - dev
-    - plugins
-    - reset_db
+    - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
+    # - {role: plugin, plugin_name: 'pulp_example', app_label: 'pulp_example'}
```
dralley: I assume this commented out line was intentional for the purpose of providing an example, but just want to check

ansible/pulp-from-source.yml
```diff
@@ -20,8 +20,8 @@
     - broker
     - lazy
     - dev
-    - plugins
-    - reset_db
+    - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
```
dralley: Hey so I found an issue with this.  If you provide the names of the plugins up-front instead of letting them be enumerated, this fails if the plugin isn't in the user's devel directory.

Specifically, I forgot to clone the pulp_file plugin, and the ansible script failed out with 

```
TASK [plugin : Install Plugin packages into core Pulp Python environment] ******
fatal: [pulp3_dev]: FAILED! => {"changed": false, "cmd": "/home/vagrant/.virtualenvs/pulp/bin/pip3 install -e /home/vagrant/devel/pulp_file", "failed": true, "msg": "\n:stderr: /home/vagrant/devel/pulp_file should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+\nYou are using pip version 8.1.2, however version 9.0.1 is available.\nYou should consider upgrading via the 'pip install --upgrade pip' command.\n"}
```

I assume we want to keep the previous workflow where you don't necessarily need to have the plugins cloned - but let me know if that assumption is invalid.

asmacdo: We have 2 choices. We can use the task to assert that the repo exists:
https://github.com/pulp/devel/blob/3.0-dev/ansible/roles/common/tasks/require_repo.yml

Or we can leave this role commented out as the default. I'm inclined to go that route. Pretty sure you could clone, uncomment and run `vagrant provision`.

Ichimonji10: I like the idea of letting the presence of pulp_file be optional, which would mean having this line of code commented out by default.

Ichimonji10:
Tested against an almost-vanilla Fedora 25 VM, the only modification being SSH configuration being in place. Installed with:

    ansible-playbook pulp-from-source.yml -i fedora-25-test-pypi,

I had to re-run it several times, whenever Ansible complained about a repository in `~/devel` being missing. Otherwise, it worked perfectly.

ansible/pulp-from-source.yml
dralley: That's fair, although it means that devs would need to uncomment/list all plugins they want to be installed themselves, which is probably fine at this point in the Pulp 3 lifecycle.  I feel like we would eventually want to go back to the old model installing everything it could find, but maybe now isn't the time to worry about it.

Thoughts?

Ichimonji10: True enough. I have several thoughts. I do like the idea of making plugin installation as seamless and easy as possible. We both want the same end goal.

I want to get to a place where we can install pulpcore (with development extras) with a single button press as soon as possible. This will be much easier to do if we start by putting a pulpcore-only playbook in front of the user by default.

asmacdo: Long story short, we were already hardcoding pulp_file into the ansible install, so for now, I think its best to move that out of the role and into the playbook. 

Id like to comment out the pulp_file install in the default playbook *and* use require_repo to fail with an easily readable message. (Probably also a small docs change). That seems like the safest default to me, and we can continue this conversation on the pulp-dev list to align our longer term ideas.

dralley: Works for me.  Let me know when those changes have been made and I will take another look

Ichimonji10: > Long story short, we were already hardcoding pulp_file into the ansible install, so for now, I think its best to move that out of the role and into the playbook.

This step?

```
TASK [plugins : Install Plugin packages into core Pulp Python environment] ********************************************
 [WARNING]: when statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ [item] | issubset(pulp_available_plugins) }}                        

skipping: [fedora-25-pulp3-pypi] => (item=pulp_file)
```

asmacdo: @Ichimonji10, yes, and that part also depends on a hardcoded list in pulp_facts. I prefer that if someone wants to modify their dev environment to install their new plugin, they should only need to change the playbook.

PR #92, opened 2017-09-19
Configure correct port for rabbitmq and remove qpid roles

werwty:


ansible/deploy-pulp3.yml
```diff
@@ -4,6 +4,10 @@
   become: true
   vars:
     ansible_python_interpreter: "/usr/bin/python3"
+    rabbitmq_config:
+      - rabbit:
+        - tcp_listeners:
+          - '"0.0.0.0"': 5672
```
mhrivnak: How about "localhost" as the default, so it's not open to the world, and it gets us ipv6?

mhrivnak: FWIW I tested with "localhost" as the value, and the pulp processes seemed to connect fine.

werwty:
@mhrivnak fixed
related documentation PR: https://github.com/pulp/pulp/pull/3156

PR #98, opened 2017-10-06
install platform and requirements before extras

asmacdo:
Despite pinning the version of DRF https://github.com/pulp/pulp/pull/3185, we ended up with a different version because the requirements are installed in the wrong order.


asmacdo:
This PR should still be merged, so if we need to pin requirements in the future, that takes precedence over extra requirements.

PR #100, opened 2017-10-11
Use pulp_user instead of 'apache'

daviddavis:


ansible/roles/dev/tasks/main.yml
```diff
@@ -1,3 +1,7 @@
+- name: create pulp user
+  user:
+    name: pulp
```
daviddavis: This is temporary until https://pulp.plan.io/issues/3065 is solved.

daviddavis:
I don't think this is a stopgap as the work here needs to be done at some point before issue 3065 is solved. I do agree that having to create a pulp user temporarily is suboptimal though so I am happy to close this out and add another checklist item to issue 3065.

ansible/roles/dev/tasks/main.yml
```diff
@@ -91,16 +95,16 @@
   file:
     path: /var/lib/pulp
     state: directory
-    owner: apache
-    group: apache
+    owner: pulp
```
asmacdo: I think this will do the trick:
s/pulp/{{ pulp_user }}/

If that works, I'm on board.

daviddavis:
@asmacdo Updated the PR based on your suggestion. I tested it out and it works well. Thanks.

PR #101, opened 2017-10-11
Fix source from which Pulp 3 roles are gotten

Ichimonji10:
The `requirements.yml` file lists which roles the `pulp-3-installer.yml`
playbook depends on. Some of the roles referenced by that file are:

* pulp.pulp3
* pulp.pulp3_db
* pulp.pulp3_systemd

Unfortunately, these roles aren't present on Ansible Galaxy. You can
prove this to yourself by going to a URL such as:
https://galaxy.ansible.com/pulp/pulp3/. A 404 is returned when getting
that page.

Fix this by directly referencing the GitHub repositories containing
these roles.

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -1,9 +1,9 @@
 # Install roles from the Ansible Galaxy
-- src: pulp.pulp3
+- src: https://github.com/pulp/ansible-pulp3.git
   name: pulp3
-- src: pulp.pulp3_db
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a local Jenkins set-up.

ammaritiz:
The Ansible Galaxy roles were disabled. I have just enabled them and you should be able to access it and download them.

Ichimonji10:
Ahh, I see. Thanks.

PR #107, opened 2018-01-09
Cleaning up the Docker provider path

dparalen:
Trying my own Pulp3 dev deployment I hit some issues.
This is my attempt to fix those in an unsorted manner.

### DIFF
--- a/ansible/roles/db/tasks/main.yml
+++ b/ansible/roles/db/tasks/main.yml
@@ -8,7 +8,8 @@
       - python3-psycopg2
 
 - name: Set up postgres data dir
-  command: postgresql-setup --initdb
+  # avoid https://code.djangoproject.com/ticket/18710
+  shell: PGSETUP_INITDB_OPTIONS='--encoding UTF-8' postgresql-setup --initdb
   args:
       creates: /var/lib/pgsql/data/base
 
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -5,6 +5,13 @@
   vars:
     repo: pulp
 
+- name: Install python SELinux bindings for the Ansible SELinux module
+  package:
+    name: "{{ item }}"
+  with_items:
+    - libselinux-python3
+    - selinux-policy
+
 - name: Disable selinux
   selinux: state=disabled
 
@@ -17,7 +24,15 @@
   notify: restart sshd
 
 - name: Install python-virtualenvwrapper
-  package: name=python-virtualenvwrapper state=latest
+  package:
+    name: "{{ item }}"
+  with_items:
+    # workaround a dependency bug of python3-virtualenvwrapper
+    - which
+    - python3-virtualenvwrapper
+    # the python2 version is required for mkvenv as mkvenv
+    # doesn't seem to respect the VIRTUALENVWRAPPER_PYTHON variable
+    - python2-virtualenvwrapper
   when: pulp_venv is defined
 
 - block:
@@ -28,9 +43,9 @@
       virtualenv: "{{ pulp_venv }}"
       virtualenv_command: /usr/bin/python3 -m venv
     with_items:
-        - common
-        - pulpcore
-        - plugin
+      - common
+      - pulpcore
+      - plugin
 
   - name: Install developer extra-requirements
     pip:
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -91,7 +91,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         override.vm.box = nil
         # Based on fedora official image, with vagrant's needs met:
         # https://github.com/rohanpm/docker-fedora-vagrant
-        d.image = 'rohanpm/fedora-vagrant:25'
+        d.image = 'rohanpm/fedora-vagrant:26'
 
         # use ssh for the sake of ansible
         d.has_ssh = true
--- a/ansible/roles/dev/files/venv.bashrc
+++ b/ansible/roles/dev/files/venv.bashrc
@@ -3,4 +3,6 @@ export WORKON_HOME=$HOME/.virtualenvs
 export PIP_VIRTUALENV_BASE=$WORKON_HOME
 export VIRTUALENV_USE_DISTRIBUTE=true
 export PIP_RESPECT_VIRTUALENV=true
+# make sure python3 is used in virtual environments
+export VIRTUALENVWRAPPER_PYTHON=/usr/bin/python3
 source /usr/bin/virtualenvwrapper.sh


ansible/pulp-from-source.yml
```diff
@@ -14,6 +14,15 @@
       package:
         name: '*'
         state: latest
+    # FIXME(milan) Docker-provider specific issues
```
asmacdo: This should probably be moved into a role.

ansible/roles/dev/tasks/main.yml
```diff
@@ -17,20 +17,39 @@
   notify: restart sshd
 
 - name: Install python-virtualenvwrapper
-  package: name=python-virtualenvwrapper state=latest
+  package:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    # FIXME(milan): a dependency bug?
+    - which
```
asmacdo: I'm surprised this is necessary. If `which` is a dep of virtualenvwrapper, it should be installed then.

ansible/roles/dev/tasks/main.yml
```diff
@@ -17,20 +17,39 @@
...
+    # doesn't seem to respect the VIRTUALENVWRAPPER_PYTHON variable
+    - python2-virtualenvwrapper
   when: pulp_venv is defined
 
 - block:
+  - name: Upgrade pip in the pulp venv
+    # NOTE(milan): besides avoiding possible issues this reduces debugging error
+    # logs by pip ranting about being out of date all the time
+    pip:
+      name: pip
```
asmacdo: should this be pip3?

ansible/pulp-from-source.yml
dparalen: Cool, which role would you recommend?

ansible/roles/dev/tasks/main.yml
dparalen: I guess the `python[23]-virtualenvwrapper` packages miss the `which` dependency:
Is `which` required in any alias?

```
(pulp) [vagrant@pulp3 pulp]$ set | grep -C3 which
+ grep --color=auto -C3 which
+ set
}
virtualenvwrapper_verify_resource ()
{
    typeset exe_path="$(command \which "$1" | (unset GREP_OPTIONS; command \grep -v "not found"))";
    if [ "$exe_path" = "" ]; then
        echo "ERROR: virtualenvwrapper could not find $1 in your path" 1>&2;
        return 1;
++ printf '\033]0;%s@%s:%s\007' vagrant pulp3 '~/devel/pulp'
```

 Create a custom function to test it:
```
(pulp) [vagrant@pulp3 pulp]$ function f () {
> typeset exe_path="$(command \which "$1" | (unset GREP_OPTIONS; command \grep -v "not found"))";
> }

(pulp) [vagrant@pulp3 pulp]$ f python
+ f python
++ command which python
++ unset GREP_OPTIONS
++ command grep -v 'not found'
++ grep -v 'not found'
++ which python
+ typeset exe_path=/home/vagrant/.virtualenvs/pulp/bin/python
++ printf '\033]0;%s@%s:%s\007' vagrant pulp3 '~/devel/pulp'
(pulp) [vagrant@pulp3 pulp]$ f foo
+ f foo
++ command which foo
++ which foo
++ unset GREP_OPTIONS
++ command grep -v 'not found'
++ grep -v 'not found'
which: no foo in (/home/vagrant/.virtualenvs/pulp/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/vagrant/.local/bin:/home/vagrant/bin)
+ typeset exe_path=
++ printf '\033]0;%s@%s:%s\007' vagrant pulp3 '~/devel/pulp'
(pulp) [vagrant@pulp3 pulp]$

```

List of deps of virtualenvwrapper:

```
pulp) [vagrant@pulp3 pulp]$ rpm -q -R python2-virtualenvwrapper
/bin/sh
config(python2-virtualenvwrapper) = 4.7.2-3.fc26
python(abi) = 2.7
python2-stevedore
python2-virtualenv
python2-virtualenv-clone
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PartialHardlinkSets) <= 4.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
(pulp) [vagrant@pulp3 pulp]$
```
`which` is actually provided by the `which` package:
```
(pulp) [vagrant@pulp3 pulp]$ dnf whatprovides '*bin/which'
Last metadata expiration check: 0:12:43 ago on Thu 11 Jan 2018 04:56:01 PM UTC.
which-2.21-2.fc26.x86_64 : Displays where a particular program in your path is located
Repo        : @System
Matched from:
Other       : *bin/which
```

ansible/roles/dev/tasks/main.yml
dparalen: I think it needn't: `pip` is being updated in the `pulp` virtual environment where it defaults to `pip3`:

```
(pulp) [vagrant@pulp3 pulp]$ which pip
~/.virtualenvs/pulp/bin/pip
(pulp) [vagrant@pulp3 pulp]$ ~/.virtualenvs/pulp/bin/pip --version
pip 9.0.1 from /home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages (python 3.6)
(pulp) [vagrant@pulp3 pulp]$
```

Ichimonji10:
I'll test this all with libvirt. I'll probably post results tomorrow.

Ichimonji10:
The following still produces functional Pulp 3 systems:

```sh
ansible-galaxy install -r requirements.yml -p roles
ansible-playbook deploy-pulp3.yml -i fedora-25-pulp-3,fedora-26-pulp-3
```

Idempotency isn't affected either, impefect as it currently is.

I can't test the `pulp-from-source.py` playbook, as it currently only works with Vagrant-backed VMs.

ansible/pulp-from-source.yml
dparalen: ...moved to the devel role... I thought it was necessary/best to do in the pre-tasks because the ansible file module might require it, but we don't actually update any SELinux file attributes (as @Ichimonji10  points out), we just disable SELinux...

ansible/roles/dev/tasks/main.yml
dparalen: No errors but every pip command in the virtual env complains about pip being out of date (when run in verbose mode).
Unless I missed some occurrence this is the first command issued in the pulp venv (creating it as a side effect).

ansible/pulp-from-source.yml
```diff
@@ -14,6 +14,15 @@
...
+    - name: Install python SELinux bindings for the Ansible file module
```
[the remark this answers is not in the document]
dparalen: my bad: wrong comment; it's the selinux module the devel role uses to disable SELinux that requires this; moved to the devel role

ansible/roles/db/tasks/main.yml
```diff
@@ -8,7 +8,8 @@
       - python3-psycopg2
 
 - name: Set up postgres data dir
-  command: postgresql-setup --initdb
+  # FIXME(milan) avoid https://code.djangoproject.com/ticket/18710
```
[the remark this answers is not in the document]
dparalen: different comment conventions I guess

ansible/roles/dev/tasks/main.yml
```diff
@@ -17,20 +17,39 @@
   notify: restart sshd
 
 - name: Install python-virtualenvwrapper
-  package: name=python-virtualenvwrapper state=latest
+  package:
+    name: "{{ item }}"
+    state: latest
```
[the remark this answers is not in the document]
dparalen: Mind please the assertion about the [dev env](https://github.com/pulp/devel/blob/3.0-dev/ansible/pulp-from-source.yml#L12)

ansible/roles/dev/tasks/main.yml
Ichimonji10: > Unless I missed some occurrence this is the first command issued in the pulp venv (creating it as a side effect).

Yes, this is the first task created after the pulp venv is created. The problem is, it will execute every subsequent time the playbook is run. A good Ansible playbook is idempotent, and this task harms the idempotency of any playbook using it.

Consider upgrading pip when the venv is first created. Something like:

```yaml
- name: Create venv
  ...
  register: result

- name: Update pip
  ...
  when: result|changed
```

ansible/roles/dev/tasks/main.yml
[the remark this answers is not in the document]
Ichimonji10: Right. And that line of code makes `latest` unnecessary. The first time this playbooks runs, the latest version of these packages will be installed. And on subsequent runs, the [dev env](https://github.com/pulp/devel/blob/3.0-dev/ansible/pulp-from-source.yml#L12) line of code will upgrade them to the latest version. Thus, this `latest` doesn't accomplish anything but to break idempotency.

Ichimonji10:
In the future, consider submitting pull requests from a branch other than `3.0-dev`. Doing so will mean that you can update your `3.0-dev` branch with the latest changes from `upstream/3.0-dev`, while leaving a pull request open. It also reduces the possibility of making mistakes, as git likes to name local branches after their `origin/*`, `upstream/*`, etc equivalent. For reading material, see something like [Git Branching - Basic Branching and Merging](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging).

ansible/roles/dev/tasks/main.yml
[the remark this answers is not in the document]
dparalen: valid point.

ansible/roles/dev/tasks/main.yml
dparalen: Any better way to create a venv than to cmd/shell?

Ichimonji10: The code here is a good example of how to create a Python 3 venv without using the `command` or `shell` modules.

dparalen: you mean to upgrade pip having installed e.g `pulp.common`?

Ichimonji10: No, I'm not talking about upgrading pip, or installing `pulp.common`. I'm answering the question above, which is how to create a venv without using the `command` or `shell` modules. Do you have a question about upgrading pip, or installing `pulp.common`?

dparalen: Sorry for not being entirely comprehensible; 
I mean what would you suggest to put instead of the ellipsis here:

```yaml
- name: Create venv
   ...
  register: result
```
...to actually create the venv

Ichimonji10: Ahh, I see now. Here's a sample playbook that creates a Python 3 venv:

```yaml
- hosts: all
  tasks:
    - name: Install requests
      pip:
        name: requests
        state: present
        virtualenv: "/home/dparalen/.venvs/test-env"
        virtualenv_command: /usr/bin/python3 -m venv
```

A couple points of interest:

* The `virtualenv` argument must be associated with an absolute path. Relative paths will not work.
* In this example, `/home/dparalen/.venvs` doesn't need to exist. More generally, the directory containing the venv being created doesn't need to exist. Intermediary directories will be created.
* The only dependency is Python 3.3+. Packages like `virtualenv`, `python2-virtualenv`, `python3-virtualenv`, and the like do not need to be installed.
* If you'd like to test this playbook, the following command should work, assuming this playbook is in a file named `site.yml`: `ansible-playbook site.yml -i localhost, --connection local`.

Ichimonji10: By the way, if you'd like to reference a virtuenv without explicitly hard-coding absolute paths into a playbook, consider a strategy like this:

```yaml
- hosts: all
  tasks:
    - name: Get information about dparalen
      getent:
        database: passwd
        key: dparalen
        split: ':'

    - name: "Set variable referencing venv"
      set_fact:
        my_venv: "{{ getent_passwd['dparalen'][4] }}/.venvs/my-venv"

    - name: Install requests
      pip:
        name: requests
        state: present
        virtualenv: "{{ my_venv }}"
        virtualenv_command: /usr/bin/python3 -m venv
```

Make sure to prefix the variable with the name of the role it's being created in. This is because variables continue to exist after their containing role has finished executing, and they can have unexpected impacts on other roles. In this case, `my_venv` should be prefixed with something.

dparalen: Thanks for the guide!
I've done some more checking and it turns out F26 already ships latest pip (9.1.z) so this isn't needed.

Cheers,
milan

dparalen:
I hope I've fixed the idempotency issues now.
I've also double checked the changes again.

dparalen:
@Ichimonji10 thanks for the review, mind merging? 
I don't have the bits to merge ;)

Ichimonji10:
I don't have write access to this repository. @asmacdo?

PR #108, opened 2018-01-18
Switches dev env to qpid

bmbouter:
Adds a new role called kombu_fixup which provides a source install of
qpid-tools and kombu.

closes #2259

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ Vagrantfile*
 .vagrant
 .dnf-cache
 ansible/*.retry
+
+# PyCharm
+.idea/
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -27,7 +27,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # to browse that repository a bit to find the latest successful Vagrant image. For example, at the
   # time of this writing, I could set this setting like this for the latest F24 image:
   # config.vm.box = "https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/CloudImages/x86_64/images/<imagename>.vagrant-libvirt.box"
-  config.vm.box = "fedora/25-cloud-base"
+  config.vm.box = "fedora/27-cloud-base"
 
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -94,3 +94,6 @@ Vagrantfile*
 .vagrant
 .dnf-cache
 ansible/*.retry
+
+# PyCharm
```
daviddavis: vim > pycharm 😄

Vagrantfile.example
```diff
@@ -39,6 +39,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision "ansible" do |ansible|
       # ansible.verbose = "vvv"
       ansible.playbook = "ansible/pulp-from-source.yml"
+      # ansible.start_at_task = "ansible task name here"
+      # ansible.extra_vars = { pulp_venv: "/home/vagrant/.virtualenvs/pulp" }
```
daviddavis: Did you mean to commit these commented out lines?

daviddavis:
Would it be easy to add like a setting (perhaps in Vagrantfile) where a user could specify which message transport he/she wants to use (default would be qpid)?

Vagrantfile.example
bmbouter: I did, I think they will be useful settings again. Really L#42 is the useful one, but then you also need L#43.

daviddavis:
Looks good 👍

bmbouter:
Yes I can add a setting that will allow the user to select qpid or rabbitmq and then default to qpid.

bmbouter:
@daviddavis I added the commented variable to Vagrantfile.example. If you uncomment that it will force it to use rabbitmq.

@dkliban I also switched it to F27.

I'm doing my final testing of both broker types now.

bmbouter:
OK I tested both environments. Both rabbitmq and qpid environments worked with F27 with the pulp_file sync commands. It's ready for independent testing and final review.

bmbouter:
Thanks for the testing! I'm also sending an email to pulp-dev explaining the change and variable to revert to rabbitmq.

daviddavis:
Awesome 👍

PR #112, opened 2018-02-15
Patch epmd.service systemd unit so it starts

dparalen:
This is to (temporarily) workaround BZ#1545840 [1]

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1545840

### DIFF
--- a/ansible/roles/broker/tasks/main.yml
+++ b/ansible/roles/broker/tasks/main.yml
@@ -15,9 +15,25 @@
   with_items:
     - rabbitmq-server
   when: use_rabbitmq == True
+  register: rabbitmq_installed
+
+- name: Patch epmd.service LimitNPROC
+  # FIXME BZ#1545840
+  lineinfile:
+    path: /usr/lib/systemd/system/epmd.service
+    state: absent
+    regexp: 'LimitNPROC=1'
+  when: rabbitmq_installed.changed == True
+  register: epmd_patched
+
+- name: Reload systemd unit files
+  systemd: daemon_reload=yes
+  when: epmd_patched.changed == True
 
 - name: Start and enable rabbitmq services
   service: name={{ item }} state=started enabled=yes
   with_items:
+    # FIXME service dependency
+    - epmd.service
     - rabbitmq-server
   when: use_rabbitmq == True


Ichimonji10:
Systemd offers the ability to patch units with separate "override" units. Rather than editing a unit that's managed by the system package manager (and causing issues when the system is updated, or making it hard to revert the patch), you can create a second unit, and any declarations in that second unit will override declarations in the default unit.

For example, let's say that I create a custom unit that fiddles with the network configuration on my host, and I want `libvirtd.service` to start after this custom unit. I can create `/etc/systemd/system/libvirtd.service.d/override.conf`, with the following contents:

```ini
[Unit]
Requires=libvirtd-networking.service
After=libvirtd-networking.service
Wants=virtlogd.service
```

I can now view the full definition of `libvirtd.service`:

```console
$ systemctl cat libvirtd.service
# /usr/lib/systemd/system/libvirtd.service
<snip!>
[Unit]
Description=Virtualization daemon
Requires=virtlogd.socket
Requires=virtlockd.socket
Before=libvirt-guests.service
After=network.target
After=dbus.service
After=iscsid.service
After=apparmor.service
After=local-fs.target
After=remote-fs.target
Documentation=man:libvirtd(8)
Documentation=https://libvirt.org\
<snip!>
# /etc/systemd/system/libvirtd.service.d/override.conf
[Unit]
Requires=libvirtd-networking.service
After=libvirtd-networking.service
Wants=virtlogd.service
```

dparalen:
Hey,
thanks for the review, I'm uncertain what you're suggesting, though: is it to drop this patch and manually create a shadow unit instead?

Cheers,
milan

Ichimonji10:
I'd suggest doing so, if reasonably possible. Doing so has several benefits:

* If the system is upgraded, then the patch won't be overwritten or cause an upgrade conflict.
* Developers can easily remove and re-add the patch.
* This is the canonical way for sysadmins to patch units.

dparalen:
Thanks for the tips!

I'm a bit lazy and would prefer the dev env deployment automation to detect&workaround this issue on my behalf.
Considering how to implement this patch, on one hand the unit file override feels cleaner, on the other hand I don't think there is much value added for the dev env.

I'd like to get more feedback from the folks though.

Cheers,
milan

dparalen:
The RabbitMQ support was dropped in  Commit https://github.com/pulp/devel/commit/4d4c13035c2f0b2a9f7fcec874a3bdbef5cfe7da  for the Pulp3 installation. The `2-master` branch was created since to hold the Pulp2 setup but the RabbitMQ support was never restored. Closing this PR.

PR #114, opened 2018-02-21
Let Pulp 3 dev be installed on non-vagrant hosts

Ichimonji10:
The Pulp 3 dev installer makes use of several local git repositories.
For example, `/home/vagrant/devel/pulp` must be present. The current
installer enforces this mandate by asserting that those directories are
present.

Don't require that these directories already exist. Instead, use them
as-is if they're present, and clone them if not. This allows Pulp dev
environments to be created on non-Vagrant hosts.

Beware that there are some nasty issues with the dev installer. For
example, the installer assumes that Ansible is connecting to the host as
user "vagrant." (This assumption is reasonable when Vagrant is used as
the VM management tool, but it's unreasonable for development hosts more
generally.) If Ansible connects as, say, root, then files and
directories owned by root will be created in `/home/vagrant`.

See: https://pulp.plan.io/issues/3318


Ichimonji10:
Successfully tested against a Fedora 26 host. After installing, I logged in and executed the following:

```sh
systemctl stop firewalld
su - vagrant
source ~/.virtualenvs/pulp/bin/activate
pulp-manager runserver 0.0.0.0:8000
```

I then tested it with Pulp Smash.

At this time, only pulpcore is installed. pulp-file and other plugins *can not be installed*, because this playbook assumes that Ansible is logging in to the target host as a user named "vagrant."

dkliban:
Tested this with my vagrant environment. Pulp works as expected after vagrant up.

PR #116, opened 2018-02-21
Drop commented-out code

Ichimonji10:
Leaving commented-out code lying around is messy:

* Commented-out code distracts from the actual, working code.
* If someone needs to know how to write a playbook, they can look at the
  Ansible documentation, or existing code in `pulp-from-source.yml` or
  `deploy-pulp3.yml`.
* If someone needs to know the past state of this playbook, they can use
  git.

### DIFF
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -33,11 +33,6 @@
     - lazy
     - dev
     - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
-    # - {role: plugin, plugin_name: 'pulp_example', app_label: 'pulp_example'}
+    # - {role: plugin, plugin_name: 'pulp_python', app_label: 'pulp_python'}
     - systemd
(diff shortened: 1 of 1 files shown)


Ichimonji10:
If documentation is a concern, we can add `README.md` files into each role.

ansible/pulp-from-source.yml
```diff
@@ -24,11 +24,5 @@
     - lazy
     - dev
     - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
-    # - {role: plugin, plugin_name: 'pulp_example', app_label: 'pulp_example'}
```
dralley: Please leave this one. It would be reasonable to change it to ```pulp_python``` though, since we're getting rid of pulp_example.

It's nice to just be able to just uncomment the plugins you want to install.

Ichimonji10: Sure. I'll do that.

PR #118, opened 2018-02-22
Drop unused "pulp_facts" module

Ichimonji10:
This module was used to find which Pulp repositories are present. It's
now useless, due to the changes implemented in
9e9b4664eb18cffd8531f765757630403d4a1b53.


Ichimonji10:
Tested with Pulp Smash on Fedora 26.

PR #121, opened 2018-02-22
Simplify django_db role

Ichimonji10:
The `django_db` role has two task files, neither of which is named
`main.yml`. This makes role invocation verbose and repetitive:

```yaml
- name: Make migrations for {{ pulp_app }}
  include_role:
    name: django_db
    tasks_from: makemigrations
  vars:
    app_label: pulp_app

- name: Run migrations
  include_role:
    name: django_db
    tasks_from: reset
```

Simplify the role, so that it can be invoked like so:

```yaml
- name: Make and run migrations for {{ pulp_app }}
  include_role:
    name: django_db
  vars:
    app_label: pulp_app
```

This is much more concise, and it works well for all current use cases.
It has a downside: If a caller wants to create several migrations and
run them all in one go, then this role doesn't allow for that. If this
situation arises, though, it can be addressed by letting the role accept
an `app_labels` var instead of an `app_label` var.

### DIFF
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -143,15 +143,11 @@
 
   become: true
 
-- include_role:
+- name: Create and run migrations for {{ pulp_app }}
+  include_role:
     name: django_db
-    tasks_from: makemigrations
   vars:
     app_label: pulp_app
-
-- include_role:
-    name: django_db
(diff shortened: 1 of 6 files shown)


Ichimonji10:
Tested against a Fedora 26 host with pulpcore and pulp-file installed.

PR #126, opened 2018-03-02
Adding sqlite binary to dev env

bmbouter:
The `dbshell` command wasn't working because it requires the binary
client. This will ensure that on dev installs at least that will work.

https://pulp.plan.io/issues/3359
re #3359

### DIFF
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -40,6 +40,12 @@
       state: present
     when: pulp_venv is defined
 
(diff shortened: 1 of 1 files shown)


ansible/roles/dev/tasks/main.yml
```diff
@@ -40,6 +40,12 @@
       state: present
     when: pulp_venv is defined
 
+  - name: Get `dbshell` command to work with sqlite3
```
daviddavis: I would just say 'Install sqlite' probably.

PR #128, opened 2018-03-07
Drop redundant variable declaration

Ichimonji10:
`group_vars/all` already declared `ansible_python_interpreter`. Don't
declare it again in the `deploy-pulp3.yml` playbook.


Ichimonji10:
Tested with Fedora 26 and 27.

PR #129, opened 2018-03-07
Don't install unnecessary package

Ichimonji10:
No custom facts are gathered, so don't install a package for gathering
custom facts.


Ichimonji10:
Tested with Fedora 26 and 27.

PR #130, opened 2018-03-07
Drop the "lazy" role

Ichimonji10:
The "lazy" role installs httpd and squid on the target host, and
configures the latter. It's called by the `pulp-from-source.yml`,
despite the fact that deferred downloading isn't supported by Pulp 3
right now, and in fact, no production web server is used by anyone at
the moment. Drop this role.


Ichimonji10:
I installed pulp_file and then tested the app with `python -m unittest discover pulp_smash.tests` on both Fedora 26 and Fedora 27.

PR #131, opened 2018-03-07
Don't install "latest" version of packages

Ichimonji10:
When installing packages, assert that necessary packages are "present,"
not "latest." The justification for installing the latest versions of
packages is that the devs want to know if Pulp 3 works with the latest
packages. However:

* Upgrading to the latest versions of packages makes the install take
  longer.
* Upgrading to the latest versions of packages makes installs
  non-idempotent.
* If a dev wants an up-to-date set of packages, it's as easy as logging
  in and executing `dnf -y update`. And the developers are likely to
  start building fresh Vagrant images daily in the near future.

### DIFF
--- a/ansible/roles/dev_tools/tasks/main.yml
+++ b/ansible/roles/dev_tools/tasks/main.yml
@@ -1,20 +1,22 @@
-# This role could be made more idempotent by requesting that packages be
-# "present," not "latest." The vim packaging issue could be solved with logic
-# that says "if the current platform is Fedora and vim-enhanced isn't installed,
-# let vim-minimal be latest."
-
-# Fedora has a packaging problem where vim-enhanced will fail to install if
-# vim-minimal is out of date.
 - block:
 
-  - name: Install latest version of vim-minimal
+  # Fedora 26 docker has a packaging problem where vim-enhanced will fail to
+  # install if vim-minimal is out of date.
+  - name: Try to install vim-enhanced
+    package:
+      name: vim-enhanced
+      state: present
+    register: result
+
+  - name: Upgrade vim-minimal if installing vim-enhanced failed
     package:
       name: vim-minimal
       state: latest
+    when: result.failed
     tags:
       - skip_ansible_lint
 
-  - name: Install latest version of several useful packages
+  - name: Install several useful packages
     package:
       name:
         - bash-completion
@@ -37,8 +39,6 @@
         - tree
         - vim-enhanced
         - yum-utils
-      state: latest
-    tags:
-      - skip_ansible_lint
+      state: present
 
   become: true
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -18,14 +18,6 @@
           This playbook requires Python 3.5+ and Ansible 2.4+. If your package
           manager ships an older version of Ansible, consider creating a
           virtualenv and installing Ansible with pip.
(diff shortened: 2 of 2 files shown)


Ichimonji10:
Tested against Fedora 26 and 27.

dparalen:
This seems reasonable but I'm a bit worried that we might be dropping compatibility issues being caught early. What would you think of a conditional execution here rather than a drop? Or postpone till we really build our own up-to-date images daily? Is this proposed with the CI speed-up in mind?

Ichimonji10:
I have two benefits in mind:

* Runs of the Pulp 3 development playbook will be faster. This is especially important for CI environments. It's also important in other cases. For example, anyone who does development work on the Ansible code benefits from faster code/test cycles. And as someone who sometimes provisions a dozen or more VMs in a single day, I greatly appreciate fast playbook executions.
* Re-runs of the Pulp 3 development playbook will be more consistent. When I'm doing development work on the Pulp 3 development playbook, I frequently run the same playbook against a single VM several times in a row, making tweaks and fixes as I go. Letting already-installed packages stay at the same version gives me confidence that changes in results are due to the development work I'm doing.

The base Jenkins VMs are re-built on a nightly basis, so merging this change should speed up jobs while still letting us catch compatibility issues quite quickly.

If y'all are especially keen on upgrading packages with every playbook run, a compromise could be to leave the upgrade pre-task in the playbook.

dparalen:
Thanks for the response!
Probably it's OK to just delegate the image being up-to-date to it's official [maintainer](https://hub.docker.com/_/fedora/) and rely on the `fedora:latest` as the stable CI/dev environment or `fedora:rawhide` as the "error preview" env. This should probably be expressed as a user preference in the `Vagrantfile` instead.

LGTM modulo my deployment [fails on vim-minimal](https://paste.fedoraproject.org/paste/T7UcCWcRZAEBs6aiFB5udg) now

Cheers,
milan

ansible/roles/dev_tools/tasks/main.yml
```diff
@@ -1,20 +1,15 @@
...
 - block:
 
-  - name: Install latest version of vim-minimal
+  - name: Install vim-minimal
     package:
       name: vim-minimal
-      state: latest
-    tags:
-      - skip_ansible_lint
+      state: present
```
dparalen: failing for me (see my comment) :-/

Ichimonji10: Adapted from an IRC conversation:

> I can't actually tell you why vim-minimal needs to be installed first. There was already code that did that when I first inherited this code base. But it didn't link to a bug tracker issue, or some other detailed explanation of why the strange installation method was used. If we're installing developer-friendly tools, the full vim seems like a good choice, not vim-minimal.

dparalen: I've just tried without the`vim-minimal` and with the regular `vim` instead of the `vim-enhanced` package and it has [this conflict](https://paste.fedoraproject.org/paste/VizxnTygNQ0deTlxGA7l7w). On the other hand, explicitly dropping `vim-minimal`here [conflicts with sudo](https://paste.fedoraproject.org/paste/6fW-Ht1Sg091ME9ZeKOFuw).

Seems this may need some workaround

Ichimonji10:
Re-tested by provisioning a pair of non-Vagrant F26 and F27 VMs, and smashing them. All was well.

PR #133, opened 2018-03-12
Don't require Python 3.5+

Ichimonji10:
It's entirely possible to install Pulp 3 on a host for which the
`ansible_python_interpreter` is older than Python 3.5. What matters is
that Python 3.5+ can be made available. For example, on
RHEL 7, the default Python interpreter is Python 2.7, but Python 3.5 can
be installed via an SCL, and Pulp can be installed into a virtualenv
that uses the SCL Python.

### DIFF
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -9,15 +9,13 @@
     pulp_user: vagrant
     use_rabbitmq: False
   pre_tasks:
-    - name: Require minimal Python and Ansible versions
+    - name: Require Ansible 2.4+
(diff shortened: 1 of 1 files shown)


Ichimonji10:
This change should have no impact on Fedora installs, and is necessary in order to enable RHEL 7 installs. This and #130 are the only PRs remaining before I can submit a PR that will enable support for RHEL 7.

asmacdo:
The original reason for requiring Python 3.5+ was some stuff in pulp_facts.py, which is apparently gone :)

PR #137, opened 2018-03-22
Removes the checkout.py script

bmbouter:
This script is no longer useful for pulp3 and it will probably stop
working soon because the data files it relies on are not being
maintained.


ammaritiz:
ok test

PR #140, opened 2018-03-27
Modify playbook to account for ansible role changes

werwty:
Changes were made in https://github.com/pulp/ansible-pulp3/pull/6
Mainly the 4 ansible roles have been smooshed into one big role.


werwty:
I know the ansible stuff isn't a priority right now, but the jenkins pulp from pypi job is broken because of this, and it was a really quick fix.

PR #141, opened 2018-04-05
Switches default broker to RabbitMQ

bmbouter:
An deadlock issue with Pulp's cancellation has been discovered with the
Qpid/kombu code. We can still test the Qpid stack as that resolves, but
we shouldn't default into a known-broken environment.


daviddavis:
Also, maybe send a note to pulp-dev telling developers if they have existing development setups to switch back to rabbitmq?

bmbouter:
I will send a note to pulp-dev.

daviddavis:
LGTM

PR #142, opened 2018-04-09
Drop reference to absent file; fix dev installs

Ichimonji10:


dralley:
This file definitely does exist, so the problem lies elsewhere.

PR #147, opened 2018-04-20
Conditionally disable IPV6 in Mongod

dekimsey:
Docker on OSX does not support ipv6 in containers. Leaving it enabled
causes provisioning to fail.

### DIFF
--- a/ansible/roles/db/tasks/main.yml
+++ b/ansible/roles/db/tasks/main.yml
@@ -13,5 +13,33 @@
       line: "nojournal = true"
   notify: restart mongod
 
+- name: Discover whether IPv6 is enabled
+  # sysctl is not installed in our dev environment
+  command: cat /proc/sys/net/ipv6/conf/lo/disable_ipv6
+  changed_when: false
+  failed_when: false
+  check_mode: false
+  register: result
+
+- block:
+
+  - name: Disable IPv6 in mongod
+    lineinfile:
+        backrefs: yes
+        dest: /etc/mongod.conf
+        regexp: "(.+)ipv6: true"
+        line: "\\1ipv6: false"
+    notify: restart mongod
+
+  - name: Remove IPv6 localhost address from mongod
+    lineinfile:
+        backrefs: yes
+        dest: /etc/mongod.conf
+        regexp: "(.+)bindIp: 127.0.0.1,::1"
+        line: "\\1bindIp: 127.0.0.1"
+    notify: restart mongod
+
+  when: result.stdout == '1'
+
 - name: Start and enable services
   service: name=mongod state=started enabled=yes


Ichimonji10:
I never deploy Pulp 2 from source, so I can't properly review this change. I can comment on some of the finer details of Ansible, though. I have a couple recommendations:

* Use a `block` to apply the `when` and `notify` to several tasks, rather than copying and pasting those lines of code. This lets the code be more DRY.
* Capture the contents of `/proc/sys/net/ipv6/conf/lo/disable_ipv6`, rather than grepping for its contents. This lets one drop the `failed_when` line of code, which makes this code more robust.
* Register task output with the `results` variable, or some other common variable. This reduces the likelihood of accidentally changing the behaviour of future tasks by defining an interesting name in their namespace. This is a real cause for concern in large Ansible playbooks, because variables persist through the entire execution of a playbook no matter where they're created, and variables created at runtime have a ridiculously high precedence.

Here's some code which should accomplish these goals:

```yaml
- name: Discover whether IPv6 is enabled
  command: sysctl --values net.ipv6.conf.lo.disable_ipv6
  changed_when: false
  check_mode: false
  register: result

- block:

  - name: Disable IPv6 in mongod
    lineinfile:
        backrefs: yes
        dest: /etc/mongod.conf
        regexp: "(.+)ipv6: true"
        line: "\\1ipv6: false"

  - name: Remove IPv6 localhost address from mongod
    lineinfile:
        backrefs: yes
        dest: /etc/mongod.conf
        regexp: "(.+)bindIp: 127.0.0.1,::1"
        line: "\\1bindIp: 127.0.0.1"

  when: result.stdout == '1'
  notify: restart mongod
```

Ichimonji10:
All that said, this code looks like it does the job.

dekimsey:
Ack, I'll tweak the Ansible. Lol, I didn't know block did that now. First time I used it when it came out didn't work well. IIRC one couldn't have a noop rescue.

dekimsey:
> Use a block to apply the when and notify to several tasks, rather than copying and pasting those lines of code. This lets the code be more DRY.

Fair enough, though notify is not a valid attirbute (at least not as of Ansible 2.3.3.0, and the [docs](http://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html) don't mention it.). I will not be implementing the notify aspect of your recommendations.

> Capture the contents of /proc/sys/net/ipv6/conf/lo/disable_ipv6, rather than grepping for its contents. This lets one drop the failed_when line of code, which makes this code more robust.

I would normally agree, however the default vagrant image doesn't include sysctl. Installing it seemed out of scope for what I was trying to accomplish.

> Register task output with the results variable, or some other common variable. This reduces the likelihood of accidentally changing the behaviour of future tasks by defining an interesting name in their namespace. This is a real cause for concern in large Ansible playbooks, because variables persist through the entire execution of a playbook no matter where they're created, and variables created at runtime have a ridiculously high precedence.

I had forgotten about this issue, our team enforces [private_role_vars](http://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-private-role-vars) on our projects nowadays. I'll be happy to change the variable being used here.

dekimsey:
`cat` works

dparalen:
Hi @dekimsey,
thanks for the PR; would you mind rebasing it on master?

Cheers,
milan

dekimsey:
@dparalen I'm looking at the PR now, it looks like there's been some attempt at committing/rebasing to it or the like.

I'm resetting and cherry-picking my changes. However, `ansible/roles/db/tasks/main.yml` no longer has any reference to mongo. And I don't see any references to it at all, has mongo been pulled from the project? If so, it would render this PR moot.

dparalen:
oh I see @dekimsey; my bad: the old `master` is now `2-master` cause Pulp3 is the `master` now ...long live the `master` ;)

Sorry for the confusion

dekimsey:
Okay, rebased onto 2-master and updated the PR branch target. Should be all cleaned up now.

PR #151, opened 2018-04-26
Revert "Adding sqlite binary to dev env"

daviddavis:
This reverts commit fe68c0a850ffe0d7361ae5956bebdc5ac886fe9b.

fixes #3505
https://pulp.plan.io/issues/3505

### DIFF
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -47,11 +47,6 @@
       - pulp_venv is defined
       - ansible_distribution == 'Fedora'
 
-  - name: Install sqlite
-    package:
-      name: sqlite
-      state: present
-
   - name: Install tools for compiling a PostgreSQL database adapter
     package:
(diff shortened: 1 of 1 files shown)


Ichimonji10:
When I say that this is "not a silver bullet for database issues," what I mean is that I'm still encountering issues such as database locks when executing tests. Here's a snippet from one traceback:

```
File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 303, in execute
    return Database.Cursor.execute(self, query, params)
django.db.utils.OperationalError: database is locked
```

dralley:
@Ichimonji10 I don't think it's intended as a fix to the sqlite issue, just part of the switch back to postgres.

Ichimonji10:
This PR states that it fixes https://pulp.plan.io/issues/3505. Is that a mis-characterization?

PR #152, opened 2018-04-30
Require Ansible 2.4+

Ichimonji10:
Ansible 2.4 or newer is required to use the new `include_*` and
`import_*` statements. The ansible-pulp3 repository uses these
constructs.

Drop the check for Python 3.5+ on the target host. This allows this
playbook to be executed against distributions like RHEL 7, which ship
python 3.5+ via SCLs. This change has already been implemented in the
`pulp-from-source.yml` playbook, which helped in making a Jenkins job
that installs Pulp 3 from source on RHEL 7.

Also see: 2b8e3e3ddc61fcc43ee3e3498a18d6f65f3a2e2e

### DIFF
--- a/ansible/deploy-pulp3.yml
+++ b/ansible/deploy-pulp3.yml
@@ -3,14 +3,12 @@
 - hosts: all
   become: true
   pre_tasks:
-    - name: Require minimal Python and Ansible versions
+    - name: Require Ansible 2.4+
       assert:
         that:
-          - "ansible_python_version|version_compare('3.5.0', operator='ge')"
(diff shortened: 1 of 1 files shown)


Ichimonji10:
The `pulp-from-source` and `deploy-pulp3` playbooks now have identical `pre_tasks`.

Ichimonji10:
This PR does one other thing, too: it drops the requirement for Python 3.5+. This is extremely useful when targeting a distribution like RHEL, which provides Python 3.5+ via SCLs.

PR #157, opened 2018-05-09
Delete unused 'scripts' and update docker devel support

dralley:
Also fix some obviously broken things in the docker box configuration


dralley:
I haven't tested the docker changes but, it's obviously broken by virtue of using the wrong name for the user reset script. 

I'd actually like to delete it entirely until some future point in time where we're ready to actually maintain / use it, but I figured I'd stick to minimal changes for now.

scripts/bootstrap-ansible.sh
```diff
@@ -1,5 +0,0 @@
-#!/bin/sh
-
-# minimal bootstrapping before kicking off ansible:
-#  install only what ansible needs to survive, or doesn't know how to do
-sudo dnf -y install python2 python2-dnf libselinux-python
```
dralley: Whether or not this is still necessary, it isn't being called run by anything on this branch and hasn't been since November 2016.

PR #162, opened 2018-06-03
Replace 'python-django' with 'python2-django1.11'

dralley:
The former has been removed from Fedora 28+
https://fedoraproject.org/wiki/Changes/Django20


dralley:
Closing because the package doesn't exist on Fedora 27

PR #163, opened 2018-06-04
Revert "Bump Fedora version"

dralley:
Reverts pulp/devel#160

Pulp 2 broken on Fedora 28 due to missing ```python-django```, ```python-rhsm``` packages.


asmacdo:
Instead of bumping to f28, I fixed the django package and install the rhsm packages from koji. 

https://github.com/pulp/devel/pull/164

PR #164, opened 2018-06-04
Update packages to match fedora repo

asmacdo:
python-django -> python2-django1.11

python-rhsm was removed, and has not yet been replaced in the upstream
repo. Instead, I added url installs to the pre-script.


dralley:
@asmacdo python2-django doesn't exist in F28 either -- there is no package which works for both F27 and F28.

asmacdo:
@dralley, I updated the comment to match what I did in the installer: python2-django1.11

asmacdo:
IMO, since we target f28 in the vagrantfile, I think we ought to get f28 working, even if it leaves f27 broken for now.

PR #168, opened 2018-06-14
Add support for CentOS

dralley:


Vagrantfile.example
```diff
@@ -37,6 +38,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   config.vm.provision "ansible" do |ansible|
+      # ansible.playbook = "ansible/configure-rhel-7.yml"
       ansible.playbook = "ansible/pulp-from-source.yml"
```
daviddavis: It's a bit unclear (to me at least) how to use CentOS. Do I uncomment line 41, do I need to also comment out line 42, etc.

It would be kind of nice if users could just change the value for `config.vm.box` to `centos/7`. Alternatively, adding instructions to README would be good.

dralley: It might be better for me to just remove the commented lines.  actually using centos (or rhel) is multi-step, you ave to run with playbook as configure-rhel-7 and then re-provision with the playbook as pulp-from-source.  It's mostly something that QE could use, or devs on very rare occasions.

Ichimonji10: Unused code is broken code. A good starting point would be to make the [`configure-rhel-7.yml`](https://github.com/pulp/devel/blob/master/ansible/configure-rhel-7.yml) playbook smarter, such that it only calls the `configure_rhel_7` role when the target host is a CentOS or RHEL 7 VM. Something like this might do it:

```yaml
# Pseudo-code!
roles:
  - name: configure_rhel_7
    when: host.ansible_distribution == 'RedHat'
```

This means that the user doesn't have to worry about un-commenting multiple lines. They just change `config.vm.box` and are done with it.

Ichimonji10:
This commit does more than just make CentOS work. It also closes the RabbitMQ port and increases the amount of RAM allocated to VMs. If those separate changes were placed into separate PRs, it would be much easier to approve those changes, and it it's the more proper thing to do anyway.

dralley:
@daviddavis @Ichimonji10 modified

ansible/configure-rhel-7.yml
```diff
@@ -4,3 +4,5 @@
     ansible_python_interpreter: /usr/bin/python2
   roles:
     - configure_rhel_7
+  when:
+    - ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
```
Ichimonji10: This syntax is broken. `when` directives can't be applied to entire plays.

ansible/pulp-from-source.yml
```diff
@@ -16,6 +16,8 @@
           This playbook requires Ansible 2.4+. Consider creating a virtualenv
           and installing Ansible with pip.
   roles:
+    - role: configure_rhel_7
+      when: ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
```
Ichimonji10: This is broken, because it will try to use the Python 3 interpreter.

dralley: There's no way to avoid that apart from taking the manual step to change ```ansible_python_interpreter``` in the ```group_vars/all``` file, which I left a note up above to do.

Pulp-CI avoids this by manually specifying that variable on the command line, so nothing changes on that front.

ansible/configure-rhel-7.yml
dralley: Fixed

ansible/pulp-from-source.yml
Ichimonji10: The ideal solution here is to set this variable in `host_vars` or `group_vars`, but since we don't version-control an inventory file, we can't use the ideal solution. Setting this varible on the command line seems like the next best option.

I agree that running two playbooks is not a great solution. It works, but it'd be far better to have just one. Given that this one playbook calls both the `configure_rhel_7` role and all the other roles, it'd probably be good to drop the `configure-rhel-7.yml` playbook in the future. The Jenkins configurations need to be updated first, though.

dralley: @Ichimonji10 Are those concerns blockers to this PR?

I'm not sure that it would be worth the effort to try to dynamically populate ansible "facts" or an inventory file, etc. based on what image is being used.  I don't suspect we'll actually need a CentOS development environment too frequently (unless it's something QE is interested in).  As far as CI goes, overriding ```ansible_python_interpreter``` on the command line is something that can be done once and and then left alone.

dralley: FWIW, I did attempt a runtime solution where, before the python parts of the ansible scripts would run, it would attempt to install the normal python3 package on CentOS so that Ansible could use the same interpreter.

Unfortunately the latest versions of Ansible don't support Python 3.4 so that didn't end up working.

Ichimonji10: > FWIW, I did attempt a runtime solution where, before the python parts of the ansible scripts would run, it would attempt to install the normal python3 package on CentOS so that Ansible could use the same interpreter.
>
> Unfortunately the latest versions of Ansible don't support Python 3.4 so that didn't end up working.

Agreed. Plus, using the Python 2.7 interpreter that ships with RHEL 7 and CentOS 7 is the most low-impact solution. I dislike using a more complicated solution (install a second interpreter) when a simpler, lower-impact solution (use the existing interpreter) is available.

Ichimonji10: > Are those concerns blockers to this PR?

I didn't list any concerns.

PR #170, opened 2018-06-20
Small devel fixes

bmbouter:
* Install jemalloc-devel which Redis needs when starting on f28 for some
  reason.
* Remove the custom RQ version we were using
* Skip errors on a place that was halting the playbook even though it
  has correct error checking.


dralley:
@bmbouter You can remove ```jemalloc```, the underlying packaging issue is being fixed.

bmbouter:
@dralley I removed the jmalloc change.

PR #171, opened 2018-06-26
Fix pulp_deb dependencies

mdellweg:
This goes hand in hand with:
https://github.com/pulp/pulp_deb/pull/45

Together they make spinning up the pulp2_dev vagrant box with pulp_deb possible again.


dralley:
With respect to ```python2-django1.11```, you cannot change it back, because that will break on Fedora 28.  More details here: https://fedoraproject.org/wiki/Changes/Django20

Relevant quote:

>The new ```python2-django1.11``` package will obsolete ```python2-django``` < 2 and ```python-django``` < 2, but it will deliberately not provide either of those. This will render packages depending on ```python(2)-django``` FTBFS (or broken runtime dependencies if they don't require Django at buildtime). This is intentional. If we don't do that, all Django packages in Fedora will just keep dragging the Legacy Python dependency chain forever. Package owners of Django applications requiring Python 2 will need to manually adjust their dependencies to use ```python2-django1.11``` or move to Python 3 only

dralley:
Disregard my earlier comment.  I will check to see when Fedora 28 packages for these deps will be available and, if it with be more than about a week, we can go ahead and roll it back.

PR #174, opened 2018-06-27
Revamp how ansible does migrations

dralley:
Problem 1:  Re-running the ansible playbook on a box with an existing Pulp installation fails because the workers have active database connections, therefore it fails when attempting to reset the db.

Problem 2:  The ```django_db``` role is run for every single plugin, resetting the database and running the migrations over and over.  Apart from being inefficient, this seems to occasionally cause problems: https://pulp.plan.io/issues/3739

* Stop the systemd services prior to doing db reset / migration steps, and restart them after.  If it can't find the services because they haven't been installed yet, ignore them.
* Let the ```plugin``` role worry about *making* the migrations, and the ```django_db``` role worry about running them.  Only run ```django_db``` role once.

### DIFF
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -24,6 +24,7 @@
     - dev
     - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
     # - {role: plugin, plugin_name: 'pulp_python', app_label: 'pulp_python'}
+    - django_db
     - systemd
     - role: dev_tools
       when: ansible_distribution == 'Fedora'
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -173,8 +173,7 @@
 
   become: true
 
-- name: Create and run migrations for {{ pulp_app }}
-  include_role:
-    name: django_db
-  vars:
-    app_label: pulp_app
+- name: Create migrations for pulp_app
+  command: >
+    {{ pulp_scl_enable_python3 }}
+    {{ pulp_venv }}/bin/pulp-manager makemigrations pulp_app
--- a/ansible/roles/django_db/tasks/main.yml
+++ b/ansible/roles/django_db/tasks/main.yml
@@ -11,26 +11,17 @@
   register: result
   failed_when: "result is failed and 'Could not find the requested service' not in result.msg"
 
-
-- block:
-
-  - name: Make migrations for {{ app_label }}
-    command: >
-      {{ pulp_scl_enable_python3 }}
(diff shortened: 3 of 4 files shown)


Ichimonji10:
The first commit does a good job improving the `django_db` role. It can be improved even further by collapsing two tasks into one, as shown below. I tested this change by running `ansible-playbook pulp-from-source.yml -i fedora-27-pulp-3-source,` twice, logging into the host, executing `systemctl stop firewalld && su - vagrant bash -c "${prefix} ~/.virtualenvs/pulp/bin/pulp-manager runserver 0.0.0.0:8000"`, and testing the system with `python -m unittest discover pulp_smash.test`.

This change isn't necessary, but I figure that since the `django_db` role is being rewritten anyway, this could be a great chance to fix up a lingering bit of silliness.

```diff
diff --git a/ansible/roles/django_db/tasks/main.yml b/ansible/roles/django_db/tasks/main.yml
index 97b7b82..65fb8ea 100644
--- a/ansible/roles/django_db/tasks/main.yml
+++ b/ansible/roles/django_db/tasks/main.yml
@@ -11,27 +11,18 @@
   register: result
   failed_when: "result is failed and 'Could not find the requested service' not in result.msg"
 
-
-- block:
-
-  - name: Make migrations for {{ app_label }}
-    command: >
-      {{ pulp_scl_enable_python3 }}
-      {{ pulp_venv }}/bin/pulp-manager makemigrations {{ app_label }}
-
-  - name: Run all migrations
-    command: >
-      {{ pulp_scl_enable_python3 }}
-      {{ pulp_venv }}/bin/pulp-manager {{ item }}
-    with_items:
-      - reset_db --noinput
-      - migrate auth --noinput
-      - migrate --noinput
-      - reset-admin-password --password admin
-
-    become: true
-    become_user: '{{ pulp_user }}'
-
+- name: Make migrations for {{ app_label }}, and run all migrations
+  command: >
+    {{ pulp_scl_enable_python3 }}
+    {{ pulp_venv }}/bin/pulp-manager {{ item }}
+  with_items:
+    - makemigrations {{ app_label }}
+    - reset_db --noinput
+    - migrate auth --noinput
+    - migrate --noinput
+    - reset-admin-password --password admin
+  become: true
+  become_user: '{{ pulp_user }}'
 
 - name: Start Pulp services
   systemd:
```

dralley:
@Ichimonji10 Doesn't the second commit already cover that?

What I can do, though, is to move "Create migrations for pulp_app" from the ```dev``` role to the ```django_db``` role, which would make sense.

Ichimonji10:
> @Ichimonji10 Doesn't the second commit already cover that?

Yes. Sorry. Ignore my comment.

PR #175, opened 2018-06-29
Create permissions for pulp_app as root

Ichimonji10:
When ansible executes tasks on a remote system, it executes those tasks
as the user it has logged in as. This behaviour can be overridden by
specifying task arguments such as `become: true`.

When Pulp 3 is being installed, migrations must be created for each of
the Django applications that comprise Pulp 3. When migrations are
created for the `pulp_app`, a permission denial occurs, because they are
being created as the default Ansible user. Fix this error by creating
migrations for `pulp_app` as root.

See: 4aa6fd29d8fa422ee01b475f387557a9a534f488

### DIFF
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -171,9 +171,9 @@
       group: '{{ pulp_user }}'
       mode: 0755
 
-  become: true
+  - name: Create migrations for pulp_app
+    command: >
+      {{ pulp_scl_enable_python3 }}
+      {{ pulp_venv }}/bin/pulp-manager makemigrations pulp_app
 
-- name: Create migrations for pulp_app
-  command: >
-    {{ pulp_scl_enable_python3 }}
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested by creating F26, F27, and RHEL 7.5 VMs, and then doing the following:

1. Create a user called "jenkins."
2. Give the jenkins user passwordless sudo access.
3. Configure Ansible to log in to the VM as the jenkins user.
4. Run roughly the following:

   ```bash
   cat >inventory <<EOF
   fedora-26-pulp-3-source
   fedora-27-pulp-3-source
   rhel-7-pulp-3-source ansible_python_interpreter=/usr/bin/python2
   EOF
   cat >rhn-credentials.yml <<EOF
   ...
   EOF
   ansible-playbook pulp-from-source.yml -i inventory -e @rhn-credentials.yml
   ```
5. Test the VMs with Pulp Smash.

Issue #176, opened 2018-07-02
pulp_user always created with defaul uid which breaks mounting (with Docker)


dparalen:
The `{{ pulp_user }}` is always created with a default `uid`. This is cool if by chance your `uid` matches the `uid` on the devel box however this wont work all the time. Imagine you create another user on your laptop and run `vagrant up...` the error might look like this:

```
# ------>%-------
TASK [dev : Install Pulp Platform packages] ************************************
failed: [pulp3_dev] (item=common) => {"changed": false, "cmd": "/home/vagrant/.virtualenvs/pulp/bin/pip install -e /home/vagrant/devel/pulp/common", "item": "common", "msg": "stdout: Obtaining file:///home/vagrant/devel/pulp/common\n    Complete output from command python setup.py egg_info:\n    running egg_info\n    creating pulpcore_common.egg-info\n    error: could not create 'pulpcore_common.egg-info': Permission denied\n    \n    ----------------------------------------\n\n:stderr: Command \"python setup.py egg_info\" failed with error code 1 in /home/vagrant/devel/pulp/common/\n"}
failed: [pulp3_dev] (item=pulpcore) => {"changed": false, "cmd": "/home/vagrant/.virtualenvs/pulp/bin/pip install -e /home/vagrant/devel/pulp/pulpcore", "item": "pulpcore", "msg": "stdout: Obtaining file:///home/vagrant/devel/pulp/pulpcore\n    Complete output from command python setup.py egg_info:\n    running egg_info\n    creating pulpcore.egg-info\n    error: could not create 'pulpcore.egg-info': Permission denied\n    \n    ----------------------------------------\n\n:stderr: Command \"python setup.py egg_info\" failed with error code 1 in /home/vagrant/devel/pulp/pulpcore/\n"}
failed: [pulp3_dev] (item=plugin) => {"changed": false, "cmd": "/home/vagrant/.virtualenvs/pulp/bin/pip install -e /home/vagrant/devel/pulp/plugin", "item": "plugin", "msg": "stdout: Obtaining file:///home/vagrant/devel/pulp/plugin\n    Complete output from command python setup.py egg_info:\n    running egg_info\n    creating pulpcore_plugin.egg-info\n    error: could not create 'pulpcore_plugin.egg-info': Permission denied\n    \n    ----------------------------------------\n\n:stderr: Command \"python setup.py egg_info\" failed with error code 1 in /home/vagrant/devel/pulp/plugin/\n"}
```

Examining the root cause, I found:
[...]

PR #177, opened 2018-07-03
usermod vagrant with host user euid

dparalen:
The vagrant user is created with a default user EUID/EGID[1]
This prevets the RW access if vagrant user EUID isn't the same as the host user EUID.

Since the uid of a user can't be changed while it has any running processes (ssh) ansible can't be used to update the euid.
The patch introduces Dockerfile that modifies the vagrant EUID/EGID build-time.

Closes: #176
[1] https://github.com/rohanpm/docker-fedora-vagrant/blob/master/Dockerfile#L30

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -90,10 +90,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
 
     pulp3_dev.vm.provider :docker do |d, override|
+
+        uid = Process.euid
+        gid = Process.egid
+
         override.vm.box = nil
         # Based on fedora official image, with vagrant's needs met:
         # https://github.com/rohanpm/docker-fedora-vagrant
-        d.image = 'rohanpm/fedora-vagrant:latest'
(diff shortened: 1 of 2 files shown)


dparalen:
Thanks @Ichimonji10!

What I noticed is the [`echo` line](https://github.com/pulp/devel/pull/177/files#diff-ebacf6f6ae4ee68078bb16454b23247dR6) also creates an image so that probably could be removed...

dparalen:
I just removed the echo line to avoid one image.
@Ichimonji10 I'm afraid it's been just me who uses docker

Ichimonji10:
Can a dev who uses Vagrant run the code in this branch and make sure it doesn't break the existing development environment? This looks sane, but I can't verify the changes by actually running them.

dparalen:
@werwty thanks for the check!
Yup, lemme update the docs!

dparalen:
@werwty just created: https://github.com/pulp/pulp/pull/3554

PR #178, opened 2018-07-03
RFC: Install plugins in dev-environment if available

mdellweg:
I am not sure, whether those playbooks are used elsewhere, but if you want to mimic the way, the vagrant dev environment included the pulp plugins, this might be the way.

### DIFF
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -22,11 +22,7 @@
     - db
     - redis
     - dev
-    - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
-    # - {role: plugin, plugin_name: 'pulp_python', app_label: 'pulp_python'}
-    # - {role: plugin, plugin_name: 'pulp_ansible', app_label: 'pulp_ansible'}
-    # - {role: plugin, plugin_name: 'pulp_docker', app_label: 'pulp_docker'}
-    # - {role: plugin, plugin_name: 'pulp_rpm', app_label: 'pulp_rpm'}
+    - plugin
     - django_db
     - systemd
     - role: dev_tools
--- a/ansible/roles/plugin/tasks/install_plugin.yml
+++ b/ansible/roles/plugin/tasks/install_plugin.yml
@@ -1,0 +1,27 @@
+- name: Get information about {{ pulp_devel_dir }}/{{ plugin_name }}
+  stat:
+    path: '{{ pulp_devel_dir }}/{{ plugin_name }}'
+  register: repo_path
+  become: true
+  become_user: '{{ pulp_user }}'
+
+- debug:
+    msg: '{{ pulp_devel_dir }}/{{ plugin_name }} does not exist, skipping!'
+  when: repo_path.stat.isdir is not defined or not repo_path.stat.isdir
+
+- block:
+
+  - name: Install plugin packages into core Pulp Python environment
+    pip:
+      name: '{{ pulp_devel_dir }}/{{ plugin_name }}'
+      extra_args: '-e'
+      executable: '{{ pulp_scl_enable_python3 }}{{ pulp_venv }}/bin/pip'
+    become: true
+    become_user: '{{ pulp_user }}'
+
+  - name: Make migrations for {{ app_label }}
+    command: >
+      {{ pulp_scl_enable_python3 }}
+      {{ pulp_venv }}/bin/pulp-manager makemigrations {{ app_label }}
+
+  when: repo_path.stat.isdir is defined and repo_path.stat.isdir
--- a/ansible/roles/plugin/tasks/main.yml
+++ b/ansible/roles/plugin/tasks/main.yml
@@ -1,27 +1,27 @@
-- name: Get information about {{ pulp_devel_dir }}/{{ plugin_name }}
-  stat:
-    path: '{{ pulp_devel_dir }}/{{ plugin_name }}'
-  register: repo_path
-  become: true
-  become_user: '{{ pulp_user }}'
+---
+- name: Look for plugin directories
+  local_action:
+    module: find
+    paths: "{{ playbook_dir }}/../../"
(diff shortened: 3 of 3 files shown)


dparalen:
@mdellweg thanks for the PR!
I'm a bit concerned about the hard-wired names, both the `pulp_file` and now `pulp_gem`.
I gave it some debugging and I think we'd better make things more dynamic. I was able to achieve that with this applied to your PR:
```diff
diff --git a/ansible/pulp-from-source.yml b/ansible/pulp-from-source.yml
index 099aa54..71a53ca 100644
--- a/ansible/pulp-from-source.yml
+++ b/ansible/pulp-from-source.yml
@@ -15,9 +15,16 @@
         msg: >
           This playbook requires Ansible 2.4+. Consider creating a virtualenv
           and installing Ansible with pip.
-    - name: Look for plugin directories
-      local_action: stat path=../../pulp_gem
-      register: pulp_gem_stat
+    - name: Look for Pulp plugins
+      local_action: find paths='{{playbook_dir + "/../../" }}' recurse=true file_type=file patterns=setup.py contains='.*pulpcore.plugin.*'
+      register: plugin_lookup
+    - name: Register Pulp plugins found
+      set_fact:
+        plugins: '{{ plugins | default([]) + [item.path | dirname | basename] }}'
+      with_items: '{{ plugin_lookup.files }}'
+      when:
+        # FIXME figure out a better solution to avoid the core match: pulp/plugin/setup.py
+        - item.path | dirname | basename != 'plugin'
   roles:
     - role: configure_rhel_7
       when: ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
@@ -25,13 +32,13 @@
     - db
     - redis
     - dev
-    - {role: plugin, plugin_name: 'pulp_file', app_label: 'pulp_file'}
-    - role: plugin
-      plugin_name: 'pulp_gem'
-      app_label: 'pulp_gem'
-      when: pulp_gem_stat.stat.isdir | default(False)
-    # - {role: plugin, plugin_name: 'pulp_python', app_label: 'pulp_python'}
-    - django_db
-    - systemd
-    - role: dev_tools
+  tasks:
+    - include_role: name=plugin
+      vars:
+        plugin_name: '{{ item }}'
+        app_label: '{{ item }}'
+      with_items: '{{ plugins }}'
+    - include_role: name=django_db
+    - include_role: name=systemd
+    - include_role: name=dev_tools
       when: ansible_distribution == 'Fedora'
```
It's a bit hairy I know but wasn't able to find a simpler dynamic solution :-/
Thoughts?

Cheers,
milan

mdellweg:
I like the dynamic approach. Let me see, whether we can improve on the lookup.
Do we assume, that all plugins are named `pulp_<plugin_slug>`?

mdellweg:
After playing around i ended up with something very similar to what you suggested.

`with_fileglob` does not work on directories, and even `with file_glob "../../*/setup.py" does not work.

dparalen:
@mdellweg yup, another inconsistency I encountered was the `excludes` on  `local_action find` wouldn't work for me at all and neither would `recurse=false` with relative paths. `with_fileglob` wouldn't work with relative paths "above" current working directory for me :-/

mdellweg:
@dparalen There is a conflicting commit adding commented lines for 'known' plugins now.
Does this mean we shall close this one? Or wait for specifically pronounced dislikes?

daviddavis:
@mdellweg I would just overwrite those changes with your changes.

daviddavis:
@mdellweg looks like [you're hitting a bug with ansible-lint](https://github.com/willthames/ansible-lint/issues/349). Any chance you could rewrite your `include_role` statements:

```yaml
- include_role: 
    name: pulp-user
```

mdellweg:
@daviddavis Now i have buried the loop one level deeper. I think this is the nicest solution.

daviddavis:
I tested this out and it worked except I have some local `pulp_...` directories that aren't Pulp 3 plugins and cause this to fail. Would you mind also adding a check for a setup.py to install_plugin.yml:

```diff
diff --git a/ansible/roles/plugin/tasks/install_plugin.yml b/ansible/roles/plugin/tasks/install_plugin.yml
index 6e70bb2..fdbc0f3 100644
--- a/ansible/roles/plugin/tasks/install_plugin.yml
+++ b/ansible/roles/plugin/tasks/install_plugin.yml
@@ -5,9 +5,14 @@
   become: true
   become_user: '{{ pulp_user }}'
 
+- name: Check that setup.py exists
+  stat:
+    path: '{{ pulp_devel_dir }}/{{ plugin_name }}/setup.py'
+  register: setup_path
+
 - debug:
-    msg: '{{ pulp_devel_dir }}/{{ plugin_name }} does not exist, skipping!'
-  when: repo_path.stat.isdir is not defined or not repo_path.stat.isdir
+    msg: '{{ pulp_devel_dir }}/{{ plugin_name }} does not exist or is not a valid plugin, skipping!'
+  when: repo_path.stat.isdir is not defined or not repo_path.stat.isdir or not setup_path.stat.exists
 
 - block:
 
@@ -24,4 +29,4 @@
       {{ pulp_scl_enable_python3 }}
       {{ pulp_venv }}/bin/pulp-manager makemigrations {{ app_label }}
 
-  when: repo_path.stat.isdir is defined and repo_path.stat.isdir
+  when: repo_path.stat.isdir is defined and repo_path.stat.isdir and setup_path.stat.exists
```

daviddavis:
> Is the existence of setup.py a good criterion to identify a pulp plugin?

It's not great but I can't think of a better one (without maybe examining file contents). I think it will suffice for now.

> Also if you like, i think we can move the whole searching into the plugin role, now.

Up to you. I am fine either way.

By the way, thanks again for working on this. 👍

mdellweg:
@daviddavis This might be the final version (if the linter likes it, too).

mdellweg:
Thanks @dparalen @daviddavis for discussions and suggestions.

PR #179, opened 2018-07-17
Sort rpm dependency list. Add python2-solv to dependencies

werwty:


dparalen:
LGTM.

For this dependency to be pulled in thru the devel environment, vagrant would have to build the RPMs.
Since we're installing from source rather than from the built RPMs we need to pull the dependencies mentioned in the spec files ourselves.

Thanks Bihan!

PR #144, opened 2018-04-10
Do not use "dev" domain

asmacdo:
Since this is a real domain, it causes problems.

PR #183, opened 2018-09-20
Do not use "dev" domain

dparalen:
Since this is a real domain, it causes problems.

A rebase and update of https://github.com/pulp/devel/pull/144


dparalen:
@daviddavis thanks, reminding you to test as requested ;)

Cheers,
milan

dralley:
However, I rebased locally for the sake of testing and it worked.  So I'm approving this PR.

PR #186, opened 2018-09-24
switches Pulp 3 to dynaconf

dkliban:


ansible/roles/dev/files/django.bashrc
```diff
@@ -1 +1,3 @@
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+export PULP_DEBUG=True
+export PULP_SECRET_KEY=$(cat /dev/urandom | tr -dc 'a-z0-9!@#$%^&*(\-_=+)' | head -c 50)
```
dparalen: wouldn't this change every time you logged in? Also mind the missing new line ;)

dkliban:
@dparalen Thanks for the review! I updated it to only generate the key once.

dparalen:
@dkliban thanks for the update!
I'll install deploy with the change briefly (in the middle of debugging another devel issue :)

ansible/roles/dev/tasks/main.yml
```diff
@@ -138,16 +138,16 @@
   - name: Install server.yaml Config file
```
dparalen: Nit(-ish): would you mind s/Install server .yaml Config file/Install server Dynaconf file/ or so?

Issue #188, opened 2018-09-25
Invalid Pulp rpm dependency requested python2-django1.11


dparalen:
This makes the `Install Pulp rpm dependencies` task fail

```
TASK [dev : Install Pulp RPM dependencies] *************************************
failed: [pulp2_dev] (item=[u'acl', u'createrepo', u'createrepo_c', u'crontabs', u'deltarpm', u'genisoimage', u'git', u'gnupg', u'gofer', u'httpd', u'kobo', u'libmodulemd', u'libselinux-python', u'm2crypto', u'mo
d_ssl', u'mod_wsgi', u'mod_xsendfile', u'nss-tools', u'openssl', u'ostree', u'policycoreutils-python', u'puppet', u'pygobject3', u'pyliblzma', u'python', u'python-blinker', u'python-celery', u'python-debian', u'
python-deltarpm', u'python-flask', u'python-gnupg', u'python-gofer', u'python-httplib2', u'python-iniparse', u'python-isodate', u'python-ldap', u'python-mongoengine', u'python-nectar', u'python-oauth2', u'python
-okaara', u'python-pycurl', u'python-pymongo', u'python-qpid', u'python-semantic_version', u'python-setuptools', u'python-twine', u'python-twisted', u'python2-debpkgr', u'python2-django1.11', u'python2-gnupg', u
'python2-lxml', u'python2-solv', u'repoview', u'rsync', u'selinux-policy', u'systemd', u'yum']) => {"changed": false, "item": ["acl", "createrepo", "createrepo_c", "crontabs", "deltarpm", "genisoimage", "git", "
gnupg", "gofer", "httpd", "kobo", "libmodulemd", "libselinux-python", "m2crypto", "mod_ssl", "mod_wsgi", "mod_xsendfile", "nss-tools", "openssl", "ostree", "policycoreutils-python", "puppet", "pygobject3", "pyli
blzma", "python", "python-blinker", "python-celery", "python-debian", "python-deltarpm", "python-flask", "python-gnupg", "python-gofer", "python-httplib2", "python-iniparse", "python-isodate", "python-ldap", "py
thon-mongoengine", "python-nectar", "python-oauth2", "python-okaara", "python-pycurl", "python-pymongo", "python-qpid", "python-semantic_version", "python-setuptools", "python-twine", "python-twisted", "python2-
[...]

PR #189, opened 2018-09-25
Fix the name of the django package to install

dparalen:
The Pulp rpm deps include python2-django1.11 which can't be installed.
Replacing with python2-django allows installation of
  python2-django-1.10.8-1.fc26.noarch

Fixes: #188


dralley:
@dparalen This will fix it on Fedora 26 and 27 but it will break it for Fedora 28+.  See: https://fedoraproject.org/wiki/Changes/Django20#Detailed_Description

dparalen:
@dralley ack; I'm dropping this as it affects just the docker which has been outdated at the source we're picking it from: last commit 2y ago in  https://github.com/rohanpm/docker-fedora-vagrant
I'm creating https://github.com/pulp/devel/pull/190 to address this.

PR #190, opened 2018-09-25
Bump Docker to use Fedora latest

dparalen:
There hasn't been much activity on the origin of our Docker image
  https://github.com/rohanpm/docker-fedora-vagrant
so I'm importing that Dockerfile locally for us to maintain

Fixes: #188


dparalen:
@dralley thanks for the review and sorry for the late answer, this one is against pulp:2-master the other one is against master

PR #191, opened 2018-09-25
Use the latest Fedora Docker container

dparalen:
Imports the origin of our container from the upstream at
https://github.com/rohanpm/docker-fedora-vagrant


dralley:
edit: Please see this first

https://github.com/pulp/devel/pull/190#pullrequestreview-159098611

dparalen:
@dralley thanks for the review!
W/r two pull requests, this PR is against pulp:master the other one is against 2-master.
My motivation was we're anyway adding custom steps but you're right; better to ping the upstream first: https://github.com/rohanpm/docker-fedora-vagrant/issues/1

Thanks!
milan

PR #192, opened 2018-10-16
Update dnf syntax for ansible 2.11

tkopecek:
Ansible is complaining about old syntax.


dralley:
Hello @tkopecek, thanks for this patch!

I ran it locally and everything looks good, except that I think one spot was missed

```
TASK [lazy : Install lazy packages] ********************************************
[DEPRECATION WARNING]: Invoking "{{ ansible_pkg_mgr }}" only once while using a
 loop via squash_actions is deprecated. Instead of using a loop to supply 
multiple items and specifying `name: {{ item }}`, please use `name: [u'squid', 
u'httpd', u'mod_ssl']` and remove the loop. This feature will be removed in 
version 2.11. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

If you could fix that, I'd be happy to merge this.

dralley:
LGTM

PR #194, opened 2018-10-19
remove broken dependency (python-django1.11)

tkopecek:
This package is not available in base repositories. Requirement should
be fulfilled by virtualenv's installations. This commit works in
symphony with https://github.com/pulp/pulp/pull/3724

Note: Doesn't help without merging https://github.com/pulp/pulp/pull/3724


tkopecek:
My fault, using different base image.

PR #198, opened 2018-11-20
Move plugin defaults

asmacdo:
defaults/ is the correct name for the default directory. default/ was not
being read by Ansible.

[noissue]


asmacdo:
Nothing. development: true will install test_requirements. The Vagrantfile.example file includes development:true to override.

https://github.com/pulp/devel/pull/197

bmbouter:
Sounds good to me. Thanks @asmacdo

PR #199, opened 2018-12-01
Add a temporary? role for createrepo_c build dependencies

dralley:
Avoids provisioning failure when pulp_rpm is one of the plugins to be installed.


bmbouter:
Yes please! This will get my pulp_rpm working out of the box with vagrant again.

PR #204, opened 2018-12-17
Add content app service

mdellweg:
Add the systemd service file, add pulp_content_app to bash helpers and
configure pulp to redirect to the content app.

https://pulp.plan.io/issues/4239
re #4239

### DIFF
--- a/ansible/roles/systemd/templates/pulp_content_app.j2
+++ b/ansible/roles/systemd/templates/pulp_content_app.j2
@@ -1,0 +1,15 @@
+[Unit]
+Description=Pulp Content App
+After=network-online.target
+Wants=network-online.target
+
+[Service]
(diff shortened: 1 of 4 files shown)


ansible/roles/systemd/templates/pulp_content_app.j2
```diff
@@ -0,0 +1,15 @@
+[Unit]
```
mdellweg: @bmbouter Is the content app meant to be run as a systemd unit?

bmbouter: That is a fine way to run it. End users could select that or other choices, whatever aiohttp supports https://aiohttp.readthedocs.io/en/stable/deployment.html

PR #205, opened 2018-12-18
Add DJANGO_SETTINGS_MODULE env variable to pulp_content_app service

gmbnomis:


gmbnomis:
@mdellweg I had to add the DJANGO_SETTINGS_MODULE environment variable to make the content app service work for me (like in the other pulp service unit definitions).  However, I wonder why you did not need it.

mdellweg:
I thought initially that it was not a django app (because it was based on aiohttp). But in the end i just forgot to make the PR for this fix. In short, yes i need it too.
See #206 .
If this gets merged first, i'm happy to rebase.

mdellweg:
Seems like the ansible linter went gaga in the meantime...

PR #207, opened 2018-12-18
Fixing ansible-lint errors

daviddavis:
A new release of ansible-lint (4.0) has caused new errors to be raised.


daviddavis:
Looks like this causes plugins not to be installed. Need to debug more tomorrow.

daviddavis:
I think this should work now. I'm having problems running vagrant up but I think they are unrelated.

daviddavis:
Finally was able to get past the problems I was hitting. I tested this and it worked.

PR #209, opened 2019-01-05
Generate settings.py from template

mdellweg:
After deleting that file in pulpcore, it needs to generated by other means.

[noissue]


mdellweg:
@asmacdo Thanks!
In general, i think it is very important, that the Vagrant development environment is fully functional at all times, because it is the first thing a new developer can be put off (and be lost forever) by. Do you think, automatic tests in that area are helpful?

PR #214, opened 2019-02-21
Use hyphens for Pulp3' system units name

mdellweg:
re #4187
https://pulp.plan.io/issues/4187

### DIFF
--- a/ansible/roles/django_db/vars/main.yaml
+++ b/ansible/roles/django_db/vars/main.yaml
@@ -1,7 +1,7 @@
 ---
 pulp_services:
-  - pulp_content_app
-  - pulp_resource_manager
-  - pulp_worker@1
-  - pulp_worker@2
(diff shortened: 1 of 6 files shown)


asmacdo:
I had suggested to @ipanova that this isn't necessary, but since we haven't deprecated this repo yet, LGTM

mdellweg:
Hey, i was only distracted for a few seconds... I wanted to ping her.

Ähm, well until pulplift is fully working, and announced to do so we must keep this working. (At least i still use it.)

daviddavis:
Did this PR not fix the problem? https://github.com/pulp/devel/pull/212

mdellweg:
It looks quite similar. Need to check.

mdellweg:
It is bit-wise exactly the same. (Should pull more often.) Closing here.

PR #216, opened 2019-03-05
Fix docker dev env provision

rohanpm:
- We shouldn't be using fedora:latest. It can change to a newer
  major version at any time, breaking the dev env.
  Right now, fedora:latest is 29 and that's broken since nightly
  repos aren't generated for F29 yet[1].

  Also, the libvirt provider for the dev-env uses F28. It seems
  a good idea for the libvirt and docker providers to be using
  the same major version of fedora.

- The hack for installing subscription manager doesn't seem needed
  any more, F28 has both subscription-manager-rhsm-certificates
  and python2-subscription-manager-rhsm

- Install dnf-utils rather than yum-utils because, on later Fedora,
  dnf-utils and yum-utils conflict.

[1] https://repos.fedorapeople.org/repos/pulp/pulp/testing/automation/2-master/stage/

### DIFF
--- a/ansible/roles/dev/tasks/main.yml
+++ b/ansible/roles/dev/tasks/main.yml
@@ -73,7 +73,7 @@
           - selinux-policy-devel
           - telnet
           - tito
-          - yum-utils
+          - dnf-utils
           - zlib-devel
 
 - name: Install gofer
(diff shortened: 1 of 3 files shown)


rohanpm:
@dralley @dparalen are you interested in having a look at this?

rohanpm:
@goosemania are you able to help me find a reviewer for this?

PR #222, opened 2019-07-24
Switch default box type to CentOS

dralley:
Modify ansible scripts and bootstrap-ansible.sh accordingly

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -27,7 +27,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # to browse that repository a bit to find the latest successful Vagrant image. For example, at the
   # time of this writing, I could set this setting like this for the latest F24 image:
   # config.vm.box = "https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/CloudImages/x86_64/images/<imagename>.vagrant-libvirt.box"
-  config.vm.box = "fedora/28-cloud-base"
+  config.vm.box =  "centos/7"
 
   # Comment out if you don't want Vagrant to add and remove entries from /etc/hosts for each VM.
   # requires the vagrant-hostmanager plugin to be installed
@@ -36,8 +36,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       config.hostmanager.manage_host = true
   end
 
+  config.vm.provision "shell", inline: "sudo yum install epel-release"
+
   # Comment this line if you would like to disable the automatic update during provisioning
-  config.vm.provision "shell", inline: "sudo dnf upgrade -y"
+  config.vm.provision "shell", inline: "sudo yum upgrade -y"
 
   # bootstrap and run with ansible
   config.vm.provision "shell", path: "scripts/bootstrap-ansible.sh"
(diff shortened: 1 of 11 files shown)


ansible/roles/dev/tasks/main.yml
```diff
@@ -97,16 +98,29 @@
       append: true
 
 - name: Install Pulp RPM dependencies
-  dnf:
+  yum:
       state: present
       name: "{{ pulp_rpm_dependencies }}"
 
 - name: Ensure homedir mode suitable for ssh keys
-  file: dest=/home/{{ ansible_env.SUDO_USER }} mode=0750
```
dralley, line 105: It's been a while (months) since I made these changes but IIRC it was actually complaining or warning about this, it wasn't just an aesthetic formatting change.

ansible/roles/pulp-certs/tasks/main.yml
```diff
@@ -35,7 +35,7 @@
 - name: Configure openssl subjectAltName
   lineinfile: dest=/etc/pki/tls/openssl.cnf
               insertafter="^\[ usr_cert \]"
-              line="subjectAltName=DNS:{{ ansible_fqdn }},DNS:{{ ansible_nodename }},DNS:{{ ansible_hostname }}"
```
dralley: I need to double check this, I don't *think* it was accidental but I don't remember why it was done.

ansible/roles/dev/tasks/main.yml
dralley: Or maybe it was the other standalone items without a name below, and I just changed others for consistency.  I don't remember anymore.

Conan-Kudo (approved):

Vagrantfile.example
```diff
@@ -72,7 +74,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         # issue and our use case, the default cpu_mode is overridden.
         domain.cpu_mode = "host-passthrough"
         domain.graphics_type = "spice"
-        domain.memory = 3076
+        domain.memory = 3096
```
mikedep333, line 77: 307*2* is the correct value for 3GB exactly. It is 3*1024.

If you want to give the kernel / (virtual) bios some memory so userspace has 3GB, 3096 sounds about right.

dralley: I'm pretty sure my motivation was never that advanced :)  I think the centos-fips branch where we first made these patches was from when we provisioned with 2gb and I increased it to a number that sounded right without actually doing the match lol.

dkliban (approved):
I tested this PR and it works great! Thank you @dralley!

scripts/bootstrap-ansible.sh
```diff
@@ -2,4 +2,4 @@
 
 # minimal bootstrapping before kicking off ansible:
 #  install only what ansible needs to survive, or doesn't know how to do
-sudo dnf -y install python2 python2-dnf libselinux-python
+sudo yum -y install python2 python2-yum libselinux-python
```
mikedep333, line 5: python2-yum does not exist.

Yum doesn't have a stable API for things like Ansible. Ansible (as intelligently as possible) wraps around the `yum` command.

dralley: Seems like the command never errored out (or had negative effects) so I never noticed.  Box seems to work fine anyways, but I'll make the change and push a new commit eventually.

PR #225, opened 2019-07-31
Use yum-config-manager not "yum config-manager"

dralley:


mikedep333 (changes requested):
/usr/bin/yum-config-manager is provided by yum-utils, but yum-utils isn't installed (or guaranteed to be installed) until later in this playbook.

mikedep333 (approved):

mikedep333 (approved):

dkliban (approved):

